PR #13, opened 2016-05-24
Feature/tests

afeena:
add some basic tests for server

### DIFF
--- a/tests/test_server.py
+++ b/tests/test_server.py
@@ -1,0 +1,47 @@
+import asyncio
+import unittest
+import json
+import aiohttp
+import builtins
+
+from unittest import mock
+
+
+class TestServer(unittest.TestCase):
+    MockedRequestHandler = None
+
+    def setUp(self):
+        p = mock.Mock()
+
+        with mock.patch('builtins.open', mock.mock_open(), create=True):
+            with mock.patch('pickle.load', p, create=True):
+                import server
+                self.MockedRequestHandler = server.HttpRequestHandler
+
+    def test_handle_request_for_dorks(self):
+        srv = self.MockedRequestHandler(debug=False, keep_alive=75)
+        transport = mock.Mock()
+        srv.connection_made(transport)
+        srv.writer = mock.Mock()
+        m = mock.Mock()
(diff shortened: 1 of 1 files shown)


tests/test_server.py
```diff
@@ -0,0 +1,105 @@
...
+import json
+
+from unittest import mock
+
+
+class TestServer(unittest.TestCase):
+    MockedRequestHandler = None
+    srv = None
+    m = None
+    m_eof = None
```
glaslos: Why do you initialize the objects here?

tests/test_server.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def setUp(self):
+        with mock.patch('builtins.open', mock.mock_open(), create=True):
+            with mock.patch('pickle.load', mock.Mock(), create=True):
+                import server
```
glaslos: Do you need to import exactly here? https://www.python.org/dev/peps/pep-0008/#imports

afeena: When I make import  at the top of the file, it tries to open file and fails, that's why I mock open function and import server after mock patching. How can I make it better?

glaslos: Hm, it shouldn't trigger https://github.com/mushorg/tanner/blob/master/server.py#L26 as it is in the scope of the class and the import should only run everything in the module scope. But I might be wrong.

glaslos: If it's too much hassle to fix it, just leave it there. This is really only a minor issue.

tests/test_server.py
afeena: It's my mistake. Should I use setUpClass() or __init__() or setUp() to initialize self variables in the test class?

glaslos: Do you have to initialize them?

afeena: I think I don't have to. I think I just can assign values to it in setUp(), correct? I tried and it works. It will be great if you clarify this.

glaslos: Correct :) I could have told you but this way you remember it better.

afeena: Thanks!

PR #21, opened 2016-05-31
Feature/track session

afeena:
Basic session tracking without storage: adding or update session, deleting old sessions


session.py
```diff
@@ -0,0 +1,23 @@
...
+        self.user_agent = data['headers']['USER-AGENT']
+        self.uuid = data['uuid']
+        self.timestamp = time.time()
+        self.count = 1
+
+    def update_session(self):
+        self.timestamp = time.time()
+        self.count += 1
+
+    def expiried(self):
```
glaslos: I think this should be `expired` :) `is_expired` might be even better.

session_manager.py
```diff
@@ -0,0 +1,43 @@
...
+    def __init__(self):
+        self.sessions = []
+
+    @asyncio.coroutine
+    def add_or_update_session(self, data):
+        #prepare the list of sessions
+        self.delete_old_sessions()
+
+        if 'peer' not in data:
+            peer = dict(ip=None, port = None)
```
glaslos: no space around equal sign of parameters.

session_manager.py
```diff
@@ -0,0 +1,43 @@
...
+            data['peer'] = peer
+
+        new_session = Session(data)
```
glaslos: only create this if we need a new session.

session_manager.py
```diff
@@ -0,0 +1,43 @@
...
+        session = self.get_session(new_session)
+        if session is None:
+            self.sessions.append(new_session)
+        else:
+            session.update_session()
+
+        print('==========EXISTING SESSIONS============')
```
glaslos: remove prints

session_manager.py
```diff
@@ -0,0 +1,43 @@
...
+        for s in self.sessions:
+            print("Session with ip", s.ip)
+        print('=======================================')
+
+    def get_session(self, session):
+        if len(self.sessions) == 0:
+            return None
+
+        for sess in self.sessions:
+            if sess.ip == session.ip:
```
glaslos: Handle when ip is `None`

PR #23, opened 2016-06-01
Feature/track and evaluate

afeena:
Add attackers' paths to the session, storing old sessions in redis storage

### DIFF
--- a/session_manager.py
+++ b/session_manager.py
@@ -1,10 +1,13 @@
 import asyncio
+import redis
+
 from session import Session
 
 
 class SessionManager:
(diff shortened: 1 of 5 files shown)


session_manager.py
```diff
@@ -34,4 +39,5 @@ def delete_old_sessions(self):
         for sess in self.sessions:
             if not sess.is_expired():
                 continue
+            self.r.set(sess.get_key(), sess.to_json())
```
glaslos: Remove the session first and then write it to redis, otherwise you can have a race condition.

PR #29, opened 2016-06-09
Feature/xss

afeena:
Basic xss, inject xss into page, get data from GET and POST requests

### DIFF
--- a/server.py
+++ b/server.py
@@ -52,10 +52,6 @@ class HttpRequestHandler(aiohttp.server.ServerHttpProtocol):
         try:
             data = json.loads(data.decode('utf-8'))
             path = data['path']
-            # dummy for wp-content
-            if re.match(r'/wp-content/.*', path):
-                m = self._make_response(msg=dict(detection={'name': 'wp-content', 'order': 1}))
-                return m
             sensor_uuid = data['uuid'] if 'uuid' in data else None
         except (TypeError, ValueError, KeyError) as e:
             print('error parsing: {}'.format(data))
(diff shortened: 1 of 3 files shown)


server.py
```diff
@@ -44,28 +48,28 @@ def _make_response(self, msg):
...
+            # dummy for wp-content
+            if re.match(r'/wp-content/.*', path):
+                m = self._make_response(msg=dict(detection={'name': 'wp-content', 'order': 1}))
+                return m
+            sensor_uuid = data['uuid'] if 'uuid' in data else None
+        except (TypeError, ValueError, KeyError) as e:
+            print('error parsing: {}'.format(data))
+            m = self._make_response(msg=type(e).__name__)
+        else:
+            session = yield from HttpRequestHandler.session_manager.add_or_update_session(data)
```
glaslos: Shouldn't that be before the wp-content match?

xss_emulator.py
```diff
@@ -0,0 +1,30 @@
...
+import mimetypes
+
+
+class XSSemulator:
+    def extract_xss_data(self, data):
+        xss = None
+        if 'post_data' in data:
+            for field, val in data['post_data'].items():
+                val = urllib.parse.unquote(val)
+                xss = re.match(r'.*<(.*)>.*', val)
```
glaslos: I think we should centralize all patterns. Can you create an issue for that?

afeena: Sure :)

server.py
afeena: Do we need store wp-content request as a separate session?

glaslos: I think we should store it as session. Right now you return before you store the session. You think we shouldn't store it as session?

afeena: I thought a bit about it and you are right. fixed

PR #30, opened 2016-06-09
Feature/enhancement

afeena:
Use uuid instead of md5 hash, add localhost phpox server and check redis connection


session.py
```diff
@@ -32,18 +34,12 @@ def is_expired(self):
     def to_json(self):
         s = dict(peer=dict(ip=self.ip, port=self.port),
                  user_agent=self.user_agent,
-                 uuid=self.uuid,
+                 sensor=self.sensor,
```
glaslos: Add the session uuid to the json

PR #39, opened 2016-06-13
tests for xss emulator

afeena:


afeena:
I wrote tests for new xss emulator class, this is not in master yet, that's why tests fail

PR #44, opened 2016-06-19
Analyze sessions in the infinite loop

afeena:


session_analyzer.py
```diff
@@ -14,14 +14,16 @@ def __init__(self):
...
+        while True:
+            session = None
+            session_key = yield
+            try:
+                session = self.r.get(session_key)
+                session = json.loads(session.decode('utf-8'))
+            except (redis.ConnectionError, TypeError) as e:
+                pass
+            print('analyze')
+            self.create_stats(session)
```
glaslos: What if you get an exception every iteration?

session_analyzer.py
```diff
@@ -14,14 +14,16 @@ def __init__(self):
...
     def analyze(self):
-        session = None
-        session_key = yield
-        try:
-            session = self.r.get(session_key)
-            session = json.loads(session.decode('utf-8'))
-        except (redis.ConnectionError, TypeError) as e:
-            pass
-        stats = self.create_stats(session)
+        while True:
```
glaslos: I would add a sleep in the loop to avoid using tons of CPU if something goes wrong.

session_analyzer.py
```diff
@@ -13,6 +14,7 @@ def __init__(self):
     @asyncio.coroutine
     def analyze(self):
         while True:
+            time.sleep(1)
```
afeena: I'm not sure it's right way, I experimented with async sleep and I have some problem with it, maybe I should rewrite this part of code and use async sleep instead of time sleep.

glaslos: What's the problem with async.sleep() ? Did you yield from it?

afeena: Yeah, I write ``yield from asyncio.sleep()`` and got the error:
```
AssertionError: yield from wasn't used with future
```

glaslos: Have a look at this: https://docs.python.org/3/library/asyncio-task.html#example-coroutine-displaying-the-current-date Which version of Python are you using?

afeena: I use Python 3.5.  I read this docs. This sample works, but when I use send() to pass the value to the generator, it fails.

glaslos:
Nice, you solve your problem :)

PR #53, opened 2016-07-06
push analyzed sessions into redis

afeena:


session_analyzer.py
```diff
@@ -9,6 +9,7 @@
 class SessionAnalyzer:
     def __init__(self):
         self.r = redis.StrictRedis(host='localhost', port=6379)
+        self.queue = []
```
glaslos: Why not use a real queue? https://docs.python.org/3/library/asyncio-queue.html

glaslos:
Well done!

PR #55, opened 2016-07-08
very basic api

afeena:
Api gives the possibility to get the list of the available SNARE uuids and get the info (sessions) of certain uuid.

### DIFF
--- a/api.py
+++ b/api.py
@@ -1,8 +1,7 @@
 import asyncio
 import redis
 import json
-import uuid
-import os
+from urllib.parse import urlparse, parse_qs
 
 
 class Api:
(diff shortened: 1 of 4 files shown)


afeena:
I think about any frameworks for working with api, maybe Flask, maybe we can migrate to it, but if we will have very limited range of api functions, maybe it will be redundant

session_manager.py
```diff
@@ -17,6 +17,8 @@ def add_or_update_session(self, raw_data):
         yield from self.delete_old_sessions()
         # handle raw data
         valid_data = self.validate_data(raw_data)
+        # push snare uuid into redis.
+        self.r.sadd(uuid.uuid3(uuid.NAMESPACE_DNS, 'snare_uuids').hex, valid_data['uuid'])
```
afeena: Maybe we can use more suitable key for snare uuids

glaslos:
I agree that Flask might be overkill. aiohttp should be good enough.

session_manager.py
glaslos: can't we just call the set `snare_ids`?

api.py
```diff
@@ -0,0 +1,37 @@
...
+
+        if path[-1] == '/':
+            path = path[:-1]
+
+        request, arg = os.path.split(path)
+
+        if request == '/api' and arg == 'stats':
+            result = yield from self.return_stats()
+        if request == '/api/stats':
+            result = yield from self.return_uuid_stats(uuid)
```
glaslos: I think you intended to do something else here.

session_manager.py
```diff
@@ -1,6 +1,6 @@
 import asyncio
 import redis
-
+import uuid
```
glaslos: not required anymore

api.py
```diff
@@ -0,0 +1,37 @@
...
+        self.r = redis.StrictRedis(host='localhost', port=6379, decode_responses=True)
+
+    @asyncio.coroutine
+    def handle_api_request(self, path):
+        result = None
+
+        if path[-1] == '/':
+            path = path[:-1]
+
+        request, arg = os.path.split(path)
```
glaslos: I'd use `/api/stats` and `/api/stats?uuid=aaaa-bbbb....` and then https://docs.python.org/3/library/urllib.parse.html or aiohttp to parse the request into components.

PR #56, opened 2016-07-21
Feature/detection

afeena:
- Add sqli detection:
 -      check path with libinjection (start subprocess with python2). 
- Move detection and emulation from server script to new script.

### DIFF
--- a/base_handler.py
+++ b/base_handler.py
@@ -1,0 +1,74 @@
+import re
+import patterns
+import urllib.parse
+import asyncio
+from asyncio.subprocess import PIPE
+import rfi_emulator
+import xss_emulator
+import lfi_emulator
+
+
+class BaseHandler:
+    # Reference patterns
+    patterns = {
+        patterns.INDEX: dict(name='index', order=1),
+        patterns.RFI_ATTACK: dict(name='rfi', order=2),
+        patterns.LFI_ATTACK: dict(name='lfi', order=2),
+        patterns.XSS_ATTACK: dict(name='xss', order=3)
+    }
+
+    def __init__(self):
+        self.rfi_emulator = rfi_emulator.RfiEmulator('/opt/tanner/')
+        self.xss_emulator = xss_emulator.XssEmulator()
+        self.lfi_emulator = lfi_emulator.LfiEmulator('/opt/tanner/')
+
+    def check_sqli(self, path):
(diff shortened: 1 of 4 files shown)


base_handler.py
```diff
@@ -0,0 +1,81 @@
...
+            xss_result = self.xss_emulator.handle(session, None, data)
+            if xss_result:
+                detection = {'name': 'xss', 'order': 2, 'payload': xss_result}
+
+        detection = dict(name='unknown', order=0)
+        # dummy for wp-content
+        if re.match(patterns.WORD_PRESS_CONTENT, path):
+            detection = {'name': 'wp-content', 'order': 1}
+
+        sqli = yield from self.check_sqli(path)
```
glaslos: I think we eventually want to pass in the parameter values here: `GET /path?foo=' OR '1'='1' --` as libinjection is evaluating SQL statements.

base_handler.py
```diff
@@ -0,0 +1,81 @@
...
+        else:
+            path = urllib.parse.unquote(path)
+            for pattern, patter_details in self.patterns.items():
+                if pattern.match(path):
+                    if detection['order'] < patter_details['order']:
+                        detection = patter_details
+        return detection
+
+    @asyncio.coroutine
+    def emulate(self, detection, session, path):
```
glaslos: With some adjustments, you could do some magic here:
```Python
getattr(self, '{}_emulator'.format(detection['name'])).handle(session, path)
```

glaslos:
I like how you separated the handlers into a separate file to keep the server.py less cluttered.
I added some comments but they are not relevant to the feature so I'm merging this. Feel free to incorporate my comments if you think they make sense.

base_handler.py
glaslos: Not sure if it's worth it thou. This makes it harder to understand the code and we only save a couple lines of code and I don't think we are going to add new handlers that often in the future.

base_handler.py
afeena: Sorry, I can't clearly understand what did you mean. Can you please explain?

glaslos: Right now you are analyzing this: `/some/path?with=parameter1&and=parameter2` But you actually want to analyze `parameter1` and `parameter2` with libinjection. libinjection evaluates SQL statements, not the entire path with parameters.

PR #57, opened 2016-07-22
all emulators are unified now

afeena:


base_handler.py
```diff
@@ -18,9 +18,11 @@ class BaseHandler:
     }
 
     def __init__(self):
-        self.rfi_emulator = rfi_emulator.RfiEmulator('/opt/tanner/')
-        self.xss_emulator = xss_emulator.XssEmulator()
-        self.lfi_emulator = lfi_emulator.LfiEmulator('/opt/tanner/')
+        self.emulators = {
```
glaslos: This is nice, better than the `getattr` magic!

glaslos:
Nice cleanup. The more often you do work like this, the cleaner the code will be and easier to work with it.

PR #64, opened 2016-08-03
catch remove None

afeena:
Catch the error when session is deleted and associated db is ```None```

### DIFF
--- a/session_manager.py
+++ b/session_manager.py
@@ -68,7 +68,7 @@ class SessionManager:
             try:
                 os.remove(sess.associated_db)
             except TypeError as e:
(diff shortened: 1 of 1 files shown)


session_manager.py
```diff
@@ -65,7 +65,10 @@ def delete_old_sessions(self):
             if not sess.is_expired():
                 continue
             #remove associated db
-            os.remove(sess.associated_db)
+            try:
+                os.remove(sess.associated_db)
+            except TypeError as e:
+                print('Cannot remove db. The db is not exists', e)
```
glaslos: The db doesn't exists :)

afeena: Thanks :)

PR #67, opened 2016-08-08
Feature/sqli table mapping

afeena:
Auto generated query map

### DIFF
--- a/sqli_emulator.py
+++ b/sqli_emulator.py
@@ -17,15 +17,41 @@ class SqliEmulator:
             os.makedirs(self.working_dir)
         db = os.path.join(self.working_dir, self.db_name)
         if not os.path.exists(db):
-           yield from self.helper.setup_db_from_config(self.working_dir, self.db_name)
+            yield from self.helper.setup_db_from_config(self.working_dir, self.db_name)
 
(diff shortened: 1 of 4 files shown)


sqli_emulator.py
```diff
@@ -17,15 +17,41 @@ def setup_db(self):
...
+                try:
+                    for row in c.execute(query):
+                        columns.append(row[1])
+                    query_map[table] = columns
+                except sqlite3.OperationalError as e:
+                    print(e)
+        return query_map
+
+    @asyncio.coroutine
+    def map_query(self, query):
```
glaslos: How often do you call this?

tests/test_sqli.py
```diff
@@ -8,7 +8,17 @@
 class SqliTest(unittest.TestCase):
     def setUp(self):
         open('/tmp/test.db', 'a').close()
+
+        @asyncio.coroutine
+        def foo():
```
glaslos: foo? :)

afeena: Naming makes me cry :)

glaslos: You are not the only one :)

PR #69, opened 2016-08-08
handle sqli in post requests

afeena:


base_handler.py
```diff
@@ -44,14 +44,25 @@ def _run_cmd(cmd):
...
-            xss_result = yield from self.emulators['xss'].handle(None, session, data)
-            if xss_result:
-                detection = {'name': 'xss', 'order': 3, 'payload': xss_result}
+        xss_result = yield from self.emulators['xss'].handle(None, session, data)
+        if xss_result:
+            detection = {'name': 'xss', 'order': 2, 'payload': xss_result}
+        else:
+            sqli_data = []
+            for (param, value) in data['post_data'].items():
+                sqli = yield from self.check_sqli(value)
```
glaslos: Shouldn't `check_sqli` be in the sqli emulator?

afeena:
Get and post data have the different structure and should be handled in different ways. Right now I don't know how make it really beautiful and clean, but maybe I can return to this code in a couple of days and change it.

PR #70, opened 2016-08-10
Use syslog instead of print

afeena:
I used syslog for the first implementation. Maybe I should rewrite it with logging module?

### DIFF
--- a/api.py
+++ b/api.py
@@ -1,12 +1,13 @@
 import asyncio
 import redis
 import json
-import syslog
+import logging
 from urllib.parse import urlparse, parse_qs
 
 
 class Api:
     def __init__(self):
+        self.logger = logging.getLogger('tanner.api.Api')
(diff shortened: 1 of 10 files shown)


glaslos:
Sorry, I mislead you, I meant to use logging and then log to syslog.

afeena:
Travis fails because of ``'dev/log'``. Maybe I'm doing something wrong with logger creation. I can mock server import in tests, but I don't like this solution. Maybe logger creation should be moved to another file or smth else... Can you give me some advice, please?

glaslos:
Right, you need a syslog service running to make it work. Maybe just go for a rotating file logger: https://docs.python.org/3.5/library/logging.handlers.html#rotatingfilehandler

afeena:
Thanks!

glaslos:
No tests? :)

PR #79, opened 2016-09-23
use values from config

afeena:
It's a raw implementation for using the values from config file . I'm a bit stuck with the architecture, I don't think that passing config as a parameter is a good idea. Maybe I can use some class for storing config, which can be accessed by any other class? I think it should have only one instance, can I use singleton? Or maybe there is another solution.

### DIFF
--- a/bin/tanner
+++ b/bin/tanner
@@ -1,35 +1,20 @@
 #!/usr/bin/python3.5
-import configparser
 import argparse
-import os
+
 from tanner import server
 from tanner.utils import logger
+from tanner import config
 
 LOGGER = logger.Logger.create_logger('tanner.log', __package__)
 
 
-def create_default_config(config):
-    if not os.path.exists('/opt/tanner/config.cfg'):
-        config = configparser.ConfigParser()
-        config['TANNER'] = {'host': '0.0.0.0', 'port': 8090}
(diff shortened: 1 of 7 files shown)


bin/tanner
```diff
@@ -1,4 +1,36 @@
...
+from tanner.utils import logger
 
-server.run_server()
\ No newline at end of file
+LOGGER = logger.Logger.create_logger('tanner.log', __package__)
+
+
+def create_default_config(config):
+    if not os.path.exists('/opt/tanner/config.cfg'):
+        config = configparser.ConfigParser()
```
glaslos: You should probably use the argument instead

bin/tanner
```diff
@@ -1,4 +1,36 @@
...
+def main():
+    config = configparser.ConfigParser()
+    create_default_config(config)
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--config", help="tanner config", default='/opt/tanner/config.cfg')
+    args = parser.parse_args()
+
+    if not os.path.exists(args.config):
+        LOGGER.warning('No config found. Use default config')
+    config.read(args.config)
```
glaslos: This probably doesn't work if I change the config path to an invalid location.

afeena:
Thanks for advice!
Sorry for the delay, I started new full-time job last week and I haven't had yet enough time to make changes. I'm going to make it on this week.

PR #83, opened 2016-11-23
Fixes

afeena:
- Prevent error when max_dorks odd number
- Update sqli_check path


setup.py
```diff
@@ -3,7 +3,7 @@
 from distutils.core import setup
 
 setup(name='Tanner',
-      version='1.0',
+      version='0.1',
```
glaslos: Can you go for a semantic versioning: major.minor.patch? So in this case 0.1.0

Issue #86, opened 2016-11-23
RFI test failing


glaslos:
FAIL: test_ftp_download (tanner.tests.test_rfi_emulation.TestRfiEmulator)

```python
Traceback (most recent call last):
  File "/home/travis/build/mushorg/tanner/tanner/tests/test_rfi_emulation.py", line 26, in test_ftp_download
    self.assertIsNotNone(data)
AssertionError: unexpectedly None
```

tanner.rfi_emulator.RfiEmulator: ERROR: Error during downloading the rfi script [Errno 101] Cannot connect to host mirror.yandex.ru:None ssl:False [Can not connect to mirror.yandex.ru:None [Network is unreachable]]

afeena:
#81 

This problem occured after new version of aiohttp has been released

afeena:
#90 #92

Issue #81, opened 2016-11-08
RFI ftp downloading fail


afeena:
It seems that new version of aiohttp fails when trying to connect over ftp.
```
tanner.rfi_emulator.RfiEmulator: ERROR: Error during downloading the rfi script [Errno 101] Cannot connect to host mirror.yandex.ru:None ssl:False [Can not connect to mirror.yandex.ru:None [Network is unreachable]]
```
We can use another library for ftp file downloading or try to solve the problem with aiohttp (if it's possible)

afeena:
#92

PR #90, opened 2016-12-01
Feature/fix paths

afeena:
fix all the paths to avoid problems with reading and setup


glaslos:
@afeena can you look into the failing test?

afeena:
Problems with lfi tests were fixed.

glaslos:
Now the only test failing is #86 which is a known issue #81

PR #94, opened 2016-12-06
Fix an error when LFI isn't properly detected

kevthehermit:
The regex will somethimes match on LFI_ATTACK but not on LFI_FILEPATH.

This causes an exception due to ```file_path = re.match(patterns.LFI_FILEPATH, path).group(1)``` being ```None```

```
2016-12-05 15:46 INFO:tanner.server.HttpRequestHandler:handle_event: Requested path /etc/passwd
2016-12-05 15:46 ERROR:tanner.server.HttpRequestHandler:log_exception: Error handling request
  File "/usr/local/lib/python3.5/dist-packages/aiohttp/server.py", line 265, in start
    yield from self.handle_request(message, payload)
  File "/usr/local/lib/python3.5/dist-packages/Tanner-0.1.0-py3.5.egg/tanner/server.py", line 75, in handle_request
    response_msg = yield from self.handle_event(data, self.redis_client)
  File "/usr/local/lib/python3.5/dist-packages/Tanner-0.1.0-py3.5.egg/tanner/server.py", line 56, in handle_event
    detection = yield from self.base_handler.handle(data, session, path)
  File "/usr/local/lib/python3.5/dist-packages/Tanner-0.1.0-py3.5.egg/tanner/emulators/base.py", line 74, in handle
    detection = yield from self.emulate(data, session, path)
  File "/usr/local/lib/python3.5/dist-packages/Tanner-0.1.0-py3.5.egg/tanner/emulators/base.py", line 68, in emulate
    emulation_result = yield from emulator.handle(path, session)
  File "/usr/local/lib/python3.5/dist-packages/Tanner-0.1.0-py3.5.egg/tanner/emulators/lfi.py", line 60, in handle
    file_path = yield from self.get_file_path(path)
  File "/usr/lib/python3.5/asyncio/coroutines.py", line 206, in coro
    res = func(*args, **kw)
  File "/usr/local/lib/python3.5/dist-packages/Tanner-0.1.0-py3.5.egg/tanner/emulators/lfi.py", line 34, in get_file_path
    file_path = re.match(patterns.LFI_FILEPATH, path).group(1)
AttributeError: 'NoneType' object has no attribute 'group'

```

### DIFF
--- a/tanner/emulators/lfi.py
+++ b/tanner/emulators/lfi.py
@@ -31,7 +31,10 @@ class LfiEmulator:
 
     @asyncio.coroutine
     def get_file_path(self, path):
-        file_path = re.match(patterns.LFI_FILEPATH, path).group(1)
+        try:
+            file_path = re.match(patterns.LFI_FILEPATH, path).group(1)
+        except Exception as e:
+            file_path = path
         file_path = os.path.normpath(os.path.join('/', file_path))
         return file_path
 


glaslos:
@afeena can you review this PR?

afeena:
@glaslos Sure.
@kevthehermit  Can you please give the example of the query, when this situation is observed?

afeena:
I think on this stage it's ok.
We will return our files only if the requested path is the same as filename (in this example, the requested path is ``/etc/passwd``, so we see the file content), in other cases they will not match and nothing will be returned.  I think it is more directory traversal that lfi, but still we can do it.
Maybe we want match only the paths which have the form ```examle.com/index.php?file=/etc/passwd```. But we can change it later.

Please, correct me, if I am wrong.

kevthehermit:
I see what you mean. and i agree Directory Traversal rather then LFI

The problem is if go to a snare instance with this

examle.com/../../../../../../etc/passwd 

snare sends the path /etc/passwd to tanner. 

Tanner regex matches for LFI attack so it goes to the LFI Emulator. When it gets to the LFI emulator the LFI_FILEPATH regex does not match so there is no group() so it fails with an unhandled exception and traceback which makes snare returns an error 500 and a nasty traceback in the console. 

The alternative would be to change the LFI attack regex or add a try except to the LFI filepath check so it returns None rather than failing with an unhandled exception. 

Hope that makes more sense.

afeena:
I got it and I totally agree that we must handle this exception, but maybe it is better to make both things: return None in any case and maybe change lfi re? So we will be sure that nothing goes wrong and TANNER will not fall.

kevthehermit:
That makes more sense, i will cancel this pull thanks :)

glaslos:
You can also just update this PR :)

afeena:
Do not cancel, I'm glad that you are doing pr =)

kevthehermit:
Haha OK, i will look at some error handling tonight and update this PR

glaslos:
@kevthehermit did you had a chance to look into this?

kevthehermit:
Sorry, back from Holidays now will have something this week

glaslos:
@kevthehermit just in case you are busy, I think we are ale to take over this and add the missing changes.

kevthehermit:
Yeah sorry I just have not had time.

afeena:
#122

PR #95, opened 2016-12-06
Log to MongoDB or Local JSON file

kevthehermit:
Adds the ability to log to a local json file or to a mongo db.

I plan to build a small web app that will visualise all the results in a similar style to kippo-graph and others.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ yarl
 redis
 asyncio_redis
 uvloop
+pymongo
--- a/tanner/config.py
+++ b/tanner/config.py
@@ -29,4 +29,6 @@ class TannerConfig():
         config['EMULATORS'] = {'root_dir': '/opt/tanner'}
         config['SQLI'] = {'db_name': 'tanner.db'}
         config['LOGGER'] = {'log_file': '/opt/tanner/tanner.log'}
+        config['MONGO'] = {'enabled': 'False', 'URI': 'mongodb://localhost'}
+        config['LOCALLOG'] = {'enabled': 'False', 'PATH': '/tmp/tanner_report.json'}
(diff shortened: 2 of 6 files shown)


glaslos:
You need to add `pymongo` to https://github.com/mushorg/tanner/blob/master/requirements.txt to make the test work again.

kevthehermit:
It seems to be upset by tthe entries in the config file.

afeena:
You can simply mock config in the server tests

```
d=dict(MONGO= {'enabled': 'False', 'URI': 'mongodb://localhost'},
       LOCALLOG= {'enabled': 'False', 'PATH': '/tmp/tanner_report.json'})
m = mock.MagicMock()
m.__getitem__.side_effect = d.__getitem__
m.__iter__.side_effect = d.__iter__
config.TannerConfig.config = m
```

kevthehermit:
Have never used mock before reading now to figure out where to put that :)

afeena:
You need to put it somewhere in setUp function of the test_server.py :)

For example:
```
class TestServer(unittest.TestCase):
    def setUp(self):
        
         MOCK (copy/paste from previous message)
        
        @asyncio.coroutine
        def choosed(client):
            return [x for x in range(10)]

```

kevthehermit:
All tests passing. Thank you very much :)

glaslos:
Nice work! Do you think we should rename `LOGGER` to `FILELOG` as he is logging lines to `tanner.log` and `LOCALLOG` to `JSONLOG` to make the naming a bit more descriptive?
I wouldn't mind merging the PR as it is and do the changes later. @afeena what do you think?

afeena:
@glaslos I totally agree :)
Great work, @kevthehermit!

PR #102, opened 2017-02-15
added requirement for libinjection

X4mp:
added instructions to setup libinjection requirements


X4mp:
Deprecated
Found python pip package

PR #116, opened 2017-03-15
check attacks based on query param values

### LINKED ISSUE #111: LFI pattern error
LFI module gives error with: `http://localhost:8080/etc/passwd`:

```
2017-03-07 15:22 INFO:tanner.server.HttpRequestHandler:handle_event: Requested path /etc/passwd
2017-03-07 15:22 ERROR:tanner.server.HttpRequestHandler:log_exception: Error handling request
  File "/usr/lib/python3/dist-packages/aiohttp/server.py", line 266, in start
    yield from self.handle_request(message, payload)
  File "/usr/local/lib/python3.5/dist-packages/Tanner-0.1.0-py3.5.egg/tanner/server.py", line 89, in handle_request
    response_msg = yield from self.handle_event(data, self.redis_client)
  File "/usr/local/lib/python3.5/dist-packages/Tanner-0.1.0-py3.5.egg/tanner/server.py", line 56, in handle_event
    detection = yield from self.base_handler.handle(data, session, path)
  File "/usr/local/lib/python3.5/dist-packages/Tanner-0.1.0-py3.5.egg/tanner/emulators/base.py", line 74, in handle
    detection = yield from self.emulate(data, session, path)
  File "/usr/local/lib/python3.5/dist-packages/Tanner-0.1.0-py3.5.egg/tanner/emulators/base.py", line 68, in emulate
    emulation_result = yield from emulator.handle(path, session)
  File "/usr/local/lib/python3.5/dist-packages/Tanner-0.1.0-py3.5.egg/tanner/emulators/lfi.py", line 58, in handle
    file_path = yield from self.get_file_path(path)
  File "/usr/lib/python3.5/asyncio/coroutines.py", line 206, in coro
    res = func(*args, **kw)
  File "/usr/local/lib/python3.5/dist-packages/Tanner-0.1.0-py3.5.egg/tanner/emulators/lfi.py", line 34, in get_file_path
    file_path = re.match(patterns.LFI_FILEPATH, path).group(1)
AttributeError: 'NoneType' object has no attribute 'group'

```

### LINKED ISSUE #113: Query check before any attack emulation
If any attack is performed via `GET` request, then it should be done using one of the query params, but the code checks for attack even if there are no query param which might lead to problems such as https://github.com/mushorg/tanner/issues/111
So for emulating a `GET` request attack, there should be a check to see whether any query param is present or NOT?

rnehra01:
fix #111
fix #113

### DIFF
--- a/tanner/emulators/base.py
+++ b/tanner/emulators/base.py
@@ -1,6 +1,7 @@
 import asyncio
 import re
 import urllib.parse
+import yarl
 
 from tanner.emulators import lfi, rfi, sqli, xss
 from tanner.utils import patterns
@@ -9,7 +10,6 @@ from tanner.utils import patterns
 class BaseHandler:
     # Reference patterns
     patterns = {
-        patterns.INDEX: dict(name='index', order=1),
         patterns.RFI_ATTACK: dict(name='rfi', order=2),
         patterns.LFI_ATTACK: dict(name='lfi', order=2),
         patterns.XSS_ATTACK: dict(name='xss', order=3)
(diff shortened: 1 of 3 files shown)


afeena:
Good work :)

Some thoughts:
If we push the value to emulators instead of the path - we don't need to extract lfi path or rfi url, we just need to check their validity. It works good and I can merge it without any changes. We can make it on the next step.

rnehra01:
@afeena Yeah we don't need to extract them, I'll try to implement.
And I think we can close this https://github.com/mushorg/tanner/pull/94.

afeena:
@rnehra01 Not exactly, I have the same error with path like `file=etc/passwd`.

rnehra01:
@afeena What should we return in this case when `file=etc/passwd`.
Since no such file exists, so we should return the requested path, what do you say?

afeena:
@rnehra01 I think we can return 404 page (no payload if file doesn't exist)

Issue #123, opened 2017-03-20
Where can I change the logging?


Philelis:
Hey,
I need a valid json format another application. I turned local logging on and saw:

> {'method': 'GET', 'peer': {'ip': '192.168.122.1', 'port': 58280}, 'path': '/images/foo.jpg', 'response_msg': b'{"response": {"message": {"detection": {"order": 1, "name": "index"}, "sess_uuid": "xxx"}}, "version": 1}', 'uuid': 'xxx, 'status': 200, 'headers': {'cache-control': 'max-age=0', 'host': '192.168.122.177', 'user-agent': 'Mozilla/5.0 (X11; Fedora; Linux x86_64; rv:52.0) Gecko/20100101 Firefox/52.0', 'accept-encoding': 'gzip, deflate', 'referer': 'http://192.168.122.177/', 'accept': '*/*', 'connection': 'keep-alive', 'cookie': 'sess_uuid=xxx', 'accept-language': 'en-GB,en;q=0.5'}, 'cookies': {'sess_uuid': 'xxx'}}

I don't want to have this b after response_msg and I need to change the output to:

> {"method": "GET","peer": {"ip": "192.168.122.1","port": 58280},"path": "/images/foo.jpg","response_msg": {"response ": {"message ": {"detection ": {"order ": 1,"name ": "index"},"sess_uuid ": "xxx"}},"version ": 1},"uuid": "xxx","status": 200,"headers": {"cache-control": "max-age=0","host": "192.168.122.177","user-agent": "Mozilla/5.0 (X11; Fedora; Linux x86_64; rv:52.0) Gecko/20100101 Firefox/52.0","accept-encoding": "gzip, deflate","referer": "http://192.168.122.177/","accept": "*/*","connection": "keep-alive","cookie": "sess_uuid=xxx","accept-language": "en-GB,en;q=0.5"},"cookies": {"sess_uuid": "xxx"}}

I did not found something like b', only in the test files. I search for that in snare and tanner.
Please help me. :)

afeena:
Hello @Philelis, the problem is here  https://github.com/mushorg/tanner/blob/master/tanner/server.py#L62
`_make_response` function return byte object.

Do you need help in fixing?

Philelis:
Yeah that would be great, how can I change this?

afeena:
You can replace L62 with `session_data['response_msg'] = response_msg.decode('utf-8')`

afeena:
[...]

PR #124, opened 2017-03-20
Make local log valid json file

afeena:
#123

### DIFF
--- a/tanner/reporting/log_local.py
+++ b/tanner/reporting/log_local.py
@@ -1,15 +1,23 @@
-import os
 import json
+import os
 
 from tanner import config
 
+
 class Reporting():
     def __init__(self):
-        # check if file exists, else create
-        pass
+        self.report_file = config.TannerConfig.get('LOCALLOG', 'PATH')
+        if not os.path.exists(self.report_file):
+            with open(self.report_file, 'w') as out:
+                json.dump([], out)
 
     def create_session(self, session_data):
-        report_file = config.TannerConfig.get('LOCALLOG', 'PATH')
-        with open(report_file, 'a') as out:
-            out.write('{0}\n'.format(session_data))
-        return '1'
+        with open(self.report_file, 'br+') as out:
+            out.seek(-1, 2)
+            out.truncate()
+
+        with open(self.report_file, 'a') as out:
+            if os.path.getsize(self.report_file) != 1:
+                out.write(",")
+            json.dump(session_data, out)
+            out.write("]")
--- a/tanner/reporting/log_local.py
+++ b/tanner/reporting/log_local.py
@@ -8,5 +8,6 @@ class Reporting:
     def create_session(session_data):
         report_file = config.TannerConfig.get('LOCALLOG', 'PATH')
         with open(report_file, 'a') as out:
-            json.dump({datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%S.%f'):session_data}, out)
+            session_data["timestamp"] = datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%S.%f')
+            json.dump(session_data, out)
             out.write('\n')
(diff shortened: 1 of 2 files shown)


tanner/reporting/log_local.py
```diff
@@ -1,15 +1,23 @@
...
-        with open(report_file, 'a') as out:
-            out.write('{0}\n'.format(session_data))
-        return '1'
+        with open(self.report_file, 'br+') as out:
+            out.seek(-1, 2)
+            out.truncate()
+
+        with open(self.report_file, 'a') as out:
+            if os.path.getsize(self.report_file) != 1:
+                out.write(",")
```
Philelis: If you want to use it with elastic change this to `out.write(",") ` to `out.write("\n") `

Philelis:
Good fix for json valid logs 👍

tanner/reporting/log_local.py
```diff
@@ -1,15 +1,23 @@
...
+            json.dump(session_data, out)
+            out.write("]")
```
Philelis: For using it with elastic delete `out.write("]")`.
Only for those they want to use elastic. This solution is complete json valid.

tanner/reporting/log_local.py
afeena: How do you think, maybe we should get delimiter from config?

afeena:
@Philelis thanks for your review :)

tanner/reporting/log_local.py
```diff
@@ -1,15 +1,23 @@
...
 from tanner import config
 
+
 class Reporting():
     def __init__(self):
-        # check if file exists, else create
-        pass
+        self.report_file = config.TannerConfig.get('LOCALLOG', 'PATH')
+        if not os.path.exists(self.report_file):
+            with open(self.report_file, 'w') as out:
```
Philelis: If someone want to use elastic:
change to `open(self.report_file, 'w')`

tanner/reporting/log_local.py
```diff
@@ -1,15 +1,23 @@
...
+                json.dump([], out)
```
Philelis: If someone want to use elastic:
delete `json.dump([], out)`

tanner/reporting/log_local.py
Philelis: I think it is not necessary, because if someone using it with things like elastic there is more to change than this.

Philelis:
I think everyone who wants to integrate this logs to elastic has now good commands what they have to change ;)

tanner/reporting/log_local.py
Philelis: If you like with timestamp:
Use this code instead of `out.write("]")`
```
 with open(self.report_file, 'br+') as out:
            out.seek(-1,2)
            out.truncate()
        with open(self.report_file, 'a') as out:
            out.write(", \"timestamp\": \""+ datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%S.%f') + "\"}]")
```

afeena:
Maybe we can add timestamp as a part of session data object, what do you think?

afeena:
I think a bit about this problem. Logs shouldn't be _completely valid json_ . If the file rather big, it's easier to work with file containing separate json objects on each line, because we can go through this file iteratively without big load on memory.

Philelis:
Your solution works, but I am not a friend of the date in front of all, I think its better to give it a name like timestamp or time and add it instead of bringing more hierarchy in the log. Do you agree with me?

afeena:
@Philelis absolutely agree :)

Philelis:
Very good solution a lot better than my one from yesterday 👍

afeena:
@Philelis Thanks :) Are you satisfied with the local logger state now? Can we merge changes? :)

Philelis:
Yeah it's perfect :)

Issue #108, opened 2017-03-02
Implement blind sqli emulation


afeena:
Expand sqli emulator features with another popular types of sqli
https://www.owasp.org/index.php/Blind_SQL_Injection

rnehra01:
Working on it.

PR #133, opened 2017-03-26
MySQLI emulator

rnehra01:
- [x] ~~Setup new database if NOT present~~
- [x] ~~Setup a new database for attacker~~
- [x] ~~Update config to take inputs like `db_password`, `choose db`~~
- [x] ~~Create abstraction for `db_helper` and `sqli`~~
- [x] ~~Identify MySQLI, handle SQLI(for this no extra function need to be written, we just need to abstract in the previous task)~~

### DIFF
--- a/tanner/emulators/mysqli.py
+++ b/tanner/emulators/mysqli.py
@@ -17,7 +17,8 @@ class MySQLIEmulator:
 
 	@asyncio.coroutine
 	def setup_db(self):
-		if not self.helper.check_db_exists(self.db_name):
+		db_exists = yield from self.helper.check_db_exists(self.db_name)
+		if not db_exists:
 			yield from self.helper.setup_db_from_config(self.db_name)
 		if self.query_map is None:
 			self.query_map = yield from self.helper.create_query_map(self.db_name)
--- a/tanner/emulators/mysqli.py
+++ b/tanner/emulators/mysqli.py
@@ -13,23 +13,18 @@ class MySQLIEmulator:
 	def __init__(self, working_dir, db_name):
 		self.db_name = db_name
 		self.helper = mysql_db_helper.MySQLDBHelper()
-		self.working_dir = os.path.join(working_dir, 'db/')
 		self.query_map = None
 
 	@asyncio.coroutine
 	def setup_db(self):
-		conn = pymysql.connect(host='localhost', user='root', password='***********')
-		cursor = conn.cursor()
-		check_DB_exists_query = 'SELECT SCHEMA_NAME FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME = \'{}\''.format(self.db_name)
-		if not cursor.execute(check_DB_exists_query):
-			yield from self.helper.setup_db_from_config(conn, self.db_name)
+		if not self.helper.check_db_exists(self.db_name):
+			yield from self.helper.setup_db_from_config(self.db_name)
 		if self.query_map is None:
-			self.query_map = yield from self.helper.create_query_map(conn, self.db_name)
-		conn.close()
+			self.query_map = yield from self.helper.create_query_map(self.db_name)
 
 
-# if __name__ == '__main__':
-# 	sqli = MySQLIEmulator('/opt/tanner', 'tanner_db')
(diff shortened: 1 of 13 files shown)


rnehra01:
@afeena Any suggestions on what and how I've done.

tanner/emulators/mysqli.py
```diff
@@ -0,0 +1,35 @@
+import asyncio
+import pymysql
+import os
+import urllib.parse
+import pylibinjection
+from asyncio.subprocess import PIPE
```
afeena: unused

afeena:
@rnehra01 For first attempt it is great :) 
I highly recommend you to move toward to high level of abstraction. I mean we shouldn't copy the same methods for every db engine. 
Also we can give the possibility to choose the db in config, for example.

rnehra01:
I have added abstraction in the list at `5` and `6` and will try to implement that soon.

tanner/emulators/mysqli.py
rnehra01: Actually imported all the things that will be used, came to my mind.

rnehra01:
@afeena take a look at `copy_db` function.

tanner/utils/mysql_db_helper.py
```diff
@@ -0,0 +1,158 @@
...
+        
+    @asyncio.coroutine
+    def setup_db_from_config(self, name=None):
+        config = yield from self.read_config()
+        if name is not None:
+            db_name = name
+        else:
+            db_name = config['name']
+               
+        conn = connect_to_db()
```
afeena: `yield from self.connect_to_db()`

tanner/utils/mysql_db_helper.py
```diff
@@ -0,0 +1,158 @@
...
+
+        conn.close()
+
+    @asyncio.coroutine
+    def copy_db(user_db, attacker_db):
+        if check_db_exists(attacker_db):
+            self.logger.info('Attacker db already exists')
+        else:
+            #create new attacker db
+            conn = connect_to_db()
```
afeena: `yield from self.connect_to_db()`

tanner/utils/mysql_db_helper.py
```diff
@@ -0,0 +1,158 @@
...
+            copy_db_cmd+= ' | '
+            copy_db_cmd+= restore_db_cmd.format(host='localhost', user='root', password='*********', db_name=attacker_db)
+            os.system(copy_db_cmd)
+
+    @asyncio.coroutine
+    def create_query_map(self,db_name, ):
+        query_map = {}
+        tables = []
+
+        conn = connect_to_db()
```
afeena: `yield from self.connect_to_db()`

rnehra01:
Will resolve them when free.

rnehra01:
@afeena resolved some of them.
Yeah, it will better to use `async library for mysql` because of its natural asynchronous nature and we have used asyncio a lot in the whole project.

glaslos:
what is the status of this PR?

rnehra01:
@glaslos I'm currently busy with my semester exams. I'll implement the remaining things after 29 Apr when my exams will end.

rnehra01:
@afeena Almost finished implementing it. Please take a look. What I was thinking is to merge `mysqli.py` file with `mysly_db_helper` as former doesn't have much functions and all can be accommodated in the latter. And same for sqlite counter part.

rnehra01:
Also can I replace sqlite db name with `tanner_db` instead of `tanner.db` so that it can match with mysql one.
Also I've prepended `attacker_` in mysql attacker db name as there is some problem in naming a database starting with integer(session id), so I'm thinking of doing same for sqlite.

rnehra01:
Fixed delete db in MySQL, port to `await-async`syntax

rnehra01:
@afeena will you merge them as a single commit or with all these, if latter let me know I'll squash some of commits named `fix` to make history look cleaner.

glaslos:
@rnehra01 no worries, we can squash from the GitHub UI. It's usually good for us to see the individual commits so we can follow your changes.

rnehra01:
Time based blind SQLI is supported by MySQL by default, #108

PR #134, opened 2017-03-28
Support strings parameter attacks

### LINKED ISSUE #130: SQLI emulator don't support string parameter
Current SQLI emulator doesn't support string parameter. If we pass a valid attack like `?username=something" UNION SELECT 1,2,3,4 --`, then it produces an unexpected error.
Though it should be a successful attack.

rnehra01:
Current SQLI emulator doesn't support string parameter. If we pass a valid attack like `?username=something" UNION SELECT 1,2,3,4 --`, then it produces an unexpected error.
Though it should be a successful attack.
With this fix, strings parameter attacks are also supported. I have put different queries for string and integer attacks.

### DIFF
--- a/tanner/emulators/sqli.py
+++ b/tanner/emulators/sqli.py
@@ -80,7 +80,7 @@ class SqliEmulator:
             if tables[0]['column']['type'] == 'INTEGER':
                 db_query = 'SELECT * from ' + tables[0]['table_name'] + ' WHERE ' + param + '=' + param_value + ';'
             else:
-                db_query = 'SELECT * from ' + tables[0]['table_name'] + ' WHERE ' + param + '= "' + param_value + '" ;'
+                db_query = 'SELECT * from ' + tables[0]['table_name'] + ' WHERE ' + param + '="' + param_value + '";'
 
         return db_query
 
--- a/tanner/emulators/sqli.py
+++ b/tanner/emulators/sqli.py
@@ -70,9 +70,17 @@ class SqliEmulator:
         db_query = None
         param = query[0][0]
         param_value = query[0][1].replace('\'', ' ')
-        tables = [k for k, v in self.query_map.items() if query[0][0] in v]
+        tables = []
+        for table, columns in self.query_map.items():
+            for column in columns: 
+                if query[0][0] == column['name']:
+                    tables.append(dict(table_name=table, column=column))
+
         if tables:
-            db_query = 'SELECT * from ' + tables[0] + ' WHERE ' + param + '=' + param_value + ';'
+            if tables[0]['column']['type'] == 'INTEGER':
+                db_query = 'SELECT * from ' + tables[0]['table_name'] + ' WHERE ' + param + '=' + param_value + ';'
+            else:
+                db_query = 'SELECT * from ' + tables[0]['table_name'] + ' WHERE ' + param + '= "' + param_value + '" ;'
 
         return db_query
 
--- a/tanner/tests/test_sqli.py
+++ b/tanner/tests/test_sqli.py
@@ -13,8 +13,11 @@ class SqliTest(unittest.TestCase):
         open('/tmp/db/test.db', 'a').close()
 
         query_map = {
-            'users': ['id', 'login', 'email', 'username', 'password', 'pass', 'log'],
-            'comments': ['comment']
+            'users': [{'name':'id', 'type':'INTEGER'}, {'name':'login', 'type':'text'},
+                      {'name':'email', 'type':'text'}, {'name':'username', 'type':'text'},
+                      {'name':'password', 'type':'text'}, {'name':'pass', 'type':'text'},
+                      {'name':'log', 'type':'text'}],
+            'comments': [{'name':'comment', 'type':'text'}]
         }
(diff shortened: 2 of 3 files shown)


glaslos:
https://github.com/mushorg/snare/pull/62#issuecomment-289708966

tanner/emulators/sqli.py
```diff
@@ -70,9 +70,17 @@ def map_query(self, query):
...
+        tables = []
+        for table, columns in self.query_map.items():
+            for column in columns: 
+                if query[0][0] == column['name']:
+                    tables.append(dict(table_name=table, column=column))
+
         if tables:
-            db_query = 'SELECT * from ' + tables[0] + ' WHERE ' + param + '=' + param_value + ';'
+            if tables[0]['column']['type'] == 'INTEGER':
+                db_query = 'SELECT * from ' + tables[0]['table_name'] + ' WHERE ' + param + '=' + param_value + ';'
```
glaslos: @afeena do you remember why we didn't used parameterized queries here?

tanner/emulators/sqli.py
```diff
@@ -70,9 +70,17 @@ def map_query(self, query):
         db_query = None
         param = query[0][0]
         param_value = query[0][1].replace('\'', ' ')
-        tables = [k for k, v in self.query_map.items() if query[0][0] in v]
+        tables = []
+        for table, columns in self.query_map.items():
+            for column in columns: 
+                if query[0][0] == column['name']:
```
glaslos: `column` is a string. I recommend running the test suit before issuing a PR.

rnehra01: I've changed that in `db_helper.py`, now it is a `dict`. Please take a look in db_helper.py.

tanner/emulators/sqli.py
afeena: @glaslos No, unfortunately, I don't. 
I suggest that it could be because of my lack of experience

tanner/emulators/sqli.py
rnehra01: I've fixed the tests.

tanner/emulators/sqli.py
glaslos: @rnehra01 we should use parameterized queries instead of concatenation.

rnehra01: @glaslos I read about `parameterized queries` and I find that they prevent `SQL Injection` but we want to do SQLI.That might be the reason you guys didn't use `parameterized queries` in the first place.

glaslos: How would an SQLi attack look like if concatenated and if parameterized?

rnehra01: If concatenated, q = `'SELECT * FROM users where name="'+some_name_var+ '";'` and an attack would make it `SELECT * FROM users where name=random" or 1=1 --";`

If parameterized using python SQLite, q=`'SELECT * FROM users where name=?'`
Now if we try to inject `random" or 1=1 --`, it won't produce anything. I'm NOT sure about how python SQLite interpret it but I guess it bypass quotes something as done by `mysql_real_escape_string` in php

Though we can use something like q = `'SELECT * FROM users where name="%s"'`

glaslos: So how do you prevent someone from getting shell access though an SQL injection in our SQLite database?

glaslos: How is `'SELECT * FROM users where name="%s"'` different from `'SELECT * FROM users where name=' + some_name_var`?

rnehra01: It's same.
Regarding shell access thing, I've no idea how to get shell access through SQLI.
Sorry, as far as I understood, this is a dummy database, so getting access to the database wouldn't create any harm and Honeypots are meant to be attacked, I guess.
Please correct if I misunderstood anything.

glaslos: http://atta.cked.me/home/sqlite3injectioncheatsheet

glaslos: What if the attacker inserts 1TB of data into the database?

rnehra01: @glaslos I think those things need the support of stacked queries, so we can disable stacked queries. And the good thing is python SQLite doesn't support stacked queries by default. So insertion or updation is NOT possible.

glaslos: So you are saying my Tanner is secure even if we allow SQL injection into the SQLite database?

rnehra01: Not sure about that, What do you suggest for tackling this?

glaslos: Why don't you tell me? :)

rnehra01: @glaslos Could you me some example of compromising tanner if there is NO write access to the database?

glaslos: Should be fine if `ATTACH DATABASE` is not supported.

rnehra01: Basically, stacked queries are not supported. So that include `ATTACH DATABASE` too.

rnehra01:
@afeena I think pr is good for merging.

PR #136, opened 2017-04-10
Feature/test improvements

afeena:
1) Changes according to #137 
2) It seems that some tests were broken - fixed. (assert raises in case when we should use asset logs, call async functions in common functions via yield from etc.)


glaslos:
Nice work. I remember form looking at other async code, it is common practice to make it possible to pass in a loop and if loop is none, use async to get the current loop.

afeena:
@glaslos Thanks :)

Issue #110, opened 2017-03-06
Deprecated Low-level Server


afeena:
The same as https://github.com/mushorg/snare/issues/55

afeena:
In process https://github.com/afeena/tanner/tree/feature/new-server

afeena:
#138

PR #138, opened 2017-04-23
Feature/new server

afeena:
#110 

New high-level server, compatible with aiohttp > 2.0

- Test will be fixed after merging with #136 
- `new_server` will be renamed into `server` before merging with master


glaslos:
Needs a rebase.

requirements.txt
```diff
@@ -1,4 +1,4 @@
-aiohttp<2.0
+aiohttp
```
glaslos: Maybe require >=2.0 ?

tanner/new_server.py
```diff
@@ -0,0 +1,108 @@
...
+    @staticmethod
+    def _make_response(msg):
+        response_message = dict(
+            version=1,
+            response=dict(message=msg)
+        )
+        return response_message
+
+    @staticmethod
+    async def default_handler(request):
```
glaslos: This is python3.6 syntax. Did you make that clear in the README?

tanner/new_server.py
```diff
@@ -0,0 +1,108 @@
...
+        except (TypeError, ValueError, KeyError) as error:
+            self.logger.error('error parsing request: %s', data)
+            response_msg = self._make_response(msg=type(error).__name__)
+        else:
+            session = await self.session_manager.add_or_update_session(
+                data, self.redis_client
+            )
+            self.logger.info('Requested path %s', path)
+            await self.dorks.extract_path(path, self.redis_client)
+            detection = await self.base_handler.handle(data, session, path)
```
glaslos: Maybe rename `handle()` into `detect` or `classify`? Should probably go into a separate PR :)

glaslos:
Looks good! Do you need another review or ready to merge?

tanner/new_server.py
afeena: I like the idea to make the names more specific :)

afeena:
@glaslos Thanks for review :) 
I want to discuss the `async/await` approach. Maybe it will be better to use `yield from` construction on this stage?

glaslos:
I don't mind the 3.6 syntax. It might cause issues on older Linux distributions. What is your opinion?

afeena:
We use `uvloop` which needs python 3.5, `async/await` is available since python 3.5.3. So I think we can migrate to new syntax without any big damage.

glaslos:
Ah, you are right! Go for it!

afeena:
Ooops, I forgot to push changes. My bad.

PR #142, opened 2017-05-17
New Sqli tests

rnehra01:


rnehra01:
I'm a little unsure about how to do it for MySQL as we will be needing Host, User, Password in travis.

afeena:
We do not need to test mysql interaction. Usually all actions with mysql db are mocked.

Don't you think that sqlite tests is connected more with sqlite helper (test especially helper's functions), not with emulator?

rnehra01:
Well, I've tried to cover all the necessary tests related to `sqlite`. You could suggest some other if you think that are needed, I'll implement them too. Or we can remove unnecessary ones.

tanner/tests/test_sqlite.py
```diff
@@ -0,0 +1,53 @@
...
+class SqliteTest(unittest.TestCase):
+	def setUp(self):
+		self.loop = asyncio.new_event_loop()
+		asyncio.set_event_loop(None)
+		self.filename = '/tmp/db/test_db'
+		os.makedirs(os.path.dirname(self.filename), exist_ok=True)
+		open('/tmp/db/test_db', 'a').close()
+		# Insert some testing data
+		conn = sqlite3.connect(self.filename)
+		self.cursor = conn.cursor()
```
afeena: Maybe we shouldn't use real db interactions in the tests? We just want to test our logic, not db interactions itself. How do you think?

afeena:
Don't worry, I just meant that tests test the functions from `sqli healper` more than in sqlite_eumulator :) We can keep them here right now and if needed we will move it in a separate file later.

rnehra01:
Yeah, you are right.
Actually, all the functions present in `SQLite` emulator are more or less DB specific.
Consider `setup_db` method, which will create a file for DB, since it is `SQLite` we can just do it using OS module but in the case of `MySQL`, need specific SQL command to do the same.
So I didn't get a way to test those without DB interaction.
That's why I thought of merging `sqlite.py` and it's helper file because they are using DB interactions.

afeena:
If you  _really_ want to test database, you can try [this](https://docs.travis-ci.com/user/database-setup/)  :)

rnehra01:
So should I remove the SQLite DB tests?

afeena:
Actually, I thought a bit and consider that we can keep it :)

PR #144, opened 2017-05-27
Command exec emulator

rnehra01:
Implemented using `docker`
Works for `GET` requests only


rnehra01:
I was thinking to use `aiodocker` library initially but that lacks documentation at present, so it is difficult to use that.
We can a TODO in the file to use `aiodocker`.

rnehra01:
I think I need some to satisfy travis.

tanner/emulators/cmd_exec.py
```diff
@@ -0,0 +1,49 @@
...
+		if not container:
+			container = self.docker_client.containers.create(image = 'ubuntu:latest',
+															 stdin_open = True, 
+															 name = container_name
+															 )
+		session.associate_env(container_name)
+		return container
+
+	async def get_cmd_exec_results(self, container, cmd):
+		container.start()
```
afeena: How will you handle the situation, if container doesn't exist (None)?

tanner/session.py
```diff
@@ -63,5 +66,12 @@ def remove_associated_db(self):
         if self.associated_db is not None and os.path.exists(self.associated_db):
             os.remove(self.associated_db)
 
+    def associate_env(self, env):
+        self.associated_env = env
+
+    @asyncio.coroutine
```
afeena: async def :)

afeena:
Look at the config
```
{
  "sudo": "required",
  "services": [
    "docker"
  ],
  "language": "python",
  "python": "3.5",
  "install": "pip install -r requirements.txt",
  "script": "nosetests -w ./tanner -vv",
  "group": "stable",
  "dist": "trusty",
  "os": "linux"
}
```
From travis [docs ](https://docs.travis-ci.com/user/docker/)

> Travis CI automatically routes builds to run on Trusty sudo: required when services: docker is configured. We do not currently support use of Docker on OS X.

I reproduced the error from the build on the local `trusty` machine. I consider it's the problem with `python 3.4`. On arch linux with `python3.6` everything is ok. If you want to use `elizabeth` on trusty - install `0.3.27` version.

tanner/emulators/cmd_exec.py
rnehra01: I'm making new container if it doesn't exist. I think you want to ask when image don't exists.

rnehra01:
I think it is complete. Regarding `POST` requests, that I would like to do in next task where I'll make a unified system for emulator so that each `post` param gets checked in each emulator.

tanner/emulators/cmd_exec.py
```diff
@@ -0,0 +1,73 @@
+import asyncio
+import docker
+import logging
+
+from tanner.config import TannerConfig
+from tanner.utils import patterns
+
+class CmdExecEmulator:
+	def __init__(self):
+		self.docker_client = docker.from_env(version='auto')
```
afeena: When docker is not installed, we got an exception which freezes all the connections 
```
    self.docker_client = docker.from_env(version='auto')
  File "/usr/local/lib/python3.5/dist-packages/docker/client.py", line 79, in from_env
    **kwargs_from_env(**kwargs))
  File "/usr/local/lib/python3.5/dist-packages/docker/client.py", line 36, in __init__
    self.api = APIClient(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/docker/api/client.py", line 141, in __init__
    self._version = self._retrieve_server_version()
  File "/usr/local/lib/python3.5/dist-packages/docker/api/client.py", line 168, in _retrieve_server_version
    'Error while fetching server API version: {0}'.format(e)
docker.errors.DockerException: Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))
```
We should provide clear manual for docker setup

rnehra01: So that will be done in documentation right? I'll do in a separate PR or should I do it here itself?

afeena: Yeah, in docks and readme. But it will be great, if you catch this exception.

rnehra01:
Sure, will do.

afeena:
Unfortunately, I can't get the expected result. I suggest that it can be because of the `;` in query:
https://github.com/mushorg/tanner/blob/master/tanner/emulators/base.py#L49
```
yarl.URL('/index.php?file=foo.txt;ls').query
<MultiDictProxy('file': 'foo.txt', 'ls': '')>
```
`ls` becomes the key of the multidict and pattern for cmd exec didn't catch the `ls`, because it checks only values

Correct me if i'm wrong :)

rnehra01:
Yeah, it's because of that part. I also checked other methods for extracting GET param such `urllib` but they did the same.

rnehra01:
And one more thing I found, the current code wouldn't support nested commands like `cd /var ; ls`. But I've found a fix for it.

afeena:
We can process cmd execution separately, as sqli :)

afeena:
Great job!
Could you please change the destination to develop and resolve the conflicts?
And one more little thing: could you please use spaces instead of tabs? All existing file use spaces, I think it will be great if all our files have the same indentation.

rnehra01:
@afeena ready for merging.

afeena:
Let's follow the pep8 recommendation and keep the spaces in the whole project 
https://www.python.org/dev/peps/pep-0008/#tabs-or-spaces :)

PR #149, opened 2017-06-08
Improve base emulator architecture

### LINKED ISSUE #148: Improve base emulator architecure
The current base emulator has become messy due a bunch of different emulators with different ways of working. Here I'm proposing a solution to handle different emulators in a standard way.
Each emulator will have a `scan` method that detects the presence of the corresponding attack in `param value` passed to it.
In the base emulator, this scan method will be called for each emulator for each `GET` and `POST` parameter.
##### BASE HANDLER
```
for each get_param:
    for each emulator:
       if (emulator.scan(param)) detection = {emulator_type, order}
//Get payload from corresponding emulator

//Similarly for post parm
```
E.g.
###### LFI
```
def scan(param):
    if (re.match(LFI.pattern, param)) return true
    else return false
```
Similary for other emulators.
In this way, we'll have a standard way in which new emulator will be formed and there won't be any need to do much change in the base emulator when new emulator will be added.
This is a basic idea. There might be some minor changes if required.

rnehra01:
With this, a new emulator can be coordinated with `base` emulator by simply adding its name to `get_emulators` or `post_emulators`. The emulator has to provide a `scan` method and this new `base emulator` will provide them all the `GET` or `POST` params vulnerable to that specific vulnerability.
Fix #148

### DIFF
--- a/tanner/emulators/base.py
+++ b/tanner/emulators/base.py
@@ -6,7 +6,6 @@ import yarl
 from tanner.emulators import lfi, rfi, sqli, xss, cmd_exec
 from tanner.utils import patterns
 
-
 class BaseHandler:
     def __init__(self, base_dir, db_name, loop=None):
         self.emulators = {
@@ -22,14 +21,6 @@ class BaseHandler:
         path = path.replace('&&', '%26%26').replace(';', '%3B')
         get_data = yarl.URL(path).query
         return get_data
-        # get_data = {}
-        # query = yarl.URL(path).query_string
-        # params = query.split('&')
-        # for param in params:
-        #     if len(param.split('=')) == 2:
-        #         param_id, param_value = param.split('=')
-        #         get_data[param_id] = param_value
-        # return get_data
 
     async def get_emulation_result(self, session, data, target_emulators):
         detection = dict(name='unknown', order=0)
@@ -37,7 +28,6 @@ class BaseHandler:
         for param_id, param_value in data.items():
             for emulator in target_emulators:
                 possible_detection = self.emulators[emulator].scan(param_value)
-                print(emulator, possible_detection)
                 if possible_detection:
                     if detection['order'] < possible_detection['order']:
                         detection = possible_detection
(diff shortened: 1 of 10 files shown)


rnehra01:
Almost done. I'll fix tests.

rnehra01:
@afeena Done.

rnehra01:
@afeena The codebase is getting large so I'm thinking that we should start putting comments in our code for each method, describing its params, function and it's return type. What do you say?

afeena:
@rnehra01 I don't mind writing docstrings, but they should be short and effective (especially for a obvious cases). Good reading is [PEP 257](https://www.python.org/dev/peps/pep-0257 ).

tanner/emulators/base.py
```diff
@@ -23,57 +15,67 @@ def __init__(self, base_dir, db_name, loop=None):
...
+        attack_params = {}
+        for param_id, param_value in data.items():
+            for emulator in target_emulators:
+                possible_detection = self.emulators[emulator].scan(param_value)
+                if possible_detection:
+                    if detection['order'] < possible_detection['order']:
+                        detection = possible_detection
+                    if emulator not in attack_params:
+                        attack_params[emulator] = []
+                    attack_params[emulator].append(dict(id= param_id, value= param_value))
```
afeena: why do you need the id of the param?

rnehra01: `id` is needed in SQLI emulator.

rnehra01:
@afeena done.

tanner/emulators/base.py
```diff
@@ -23,57 +15,66 @@ def __init__(self, base_dir, db_name, loop=None):
...
+        encodings = [('&&', '%26%26'), (';', '%3B')] 
+        for value, encoded_value in encodings:
+            path = path.replace(value, encoded_value)
+        get_data = yarl.URL(path).query
+        return get_data
+
+    async def get_emulation_result(self, session, data, target_emulators):
+        """
+        Return emulation result for the vulnerabilty of highest order
+        :param session: Current active session
```
afeena: I think in this case params of the function is completely understandable from the context. Do we really need to write that session is session? :) I think we should write more self-documented code than more docstrings. How do you think?

rnehra01: Well, self-documented code is always the first choice. But the thing is we've been working with a code so we feel it is unnecessary, but when I saw the code first time, it is difficult to go through the whole method and figure out and you just need to know what the method takes input and how does output look like. So I thought of writing doc strings. If you think it isn't needed at this stage, we can remove them but there is no harm in keeping them.

afeena: Yes, you are right. I don't want to remove docstrings. I completely understand your point of view, I read lots of source code and know this feeling. But when I read the new code, the most important thing for me is input/output type. I saw the string "Current active session" and asked myself: "is it string, is it number or is it other type?". But if I see the string "Session class object" it makes difference. I don't need to find the place where the current session is created, to understand it's type. I 100% sure that I'll find all the information in session class.

afeena:
Good work :)

Issue #150, opened 2017-06-11
Add support of cookies for attacks


rnehra01:
With the addition of functionality of cookies(​ #112​ ), we can now check attacks 
in cookies too.

rnehra01:
#152 closes this.

PR #152, opened 2017-06-14
Add Cookie support for attacks

rnehra01:
@afeena Is it fine, or should I call `handle_cookies` in `emulate` method. I think this is better.
Tests will be fixed based on that.

### DIFF
--- a/tanner/emulators/base.py
+++ b/tanner/emulators/base.py
@@ -17,6 +17,7 @@ class BaseHandler:
         }
         self.get_emulators = ['sqli', 'rfi', 'lfi', 'xss', 'cmd_exec']
         self.post_emulators = ['sqli', 'rfi', 'lfi', 'xss', 'cmd_exec']
+        self.cookie_emulators = ['sqli']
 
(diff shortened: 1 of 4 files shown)


afeena:
@rnehra01 I don't think changing `emulate` method will be better. Keep this solution :)

rnehra01:
There is one thing that might cause problems i.e presence of  `=` in cookies' values. Because we are extracting cookie data using `split('=')`.
One solution to it is that we should accept cookies with some sort of encoding, for simplicity `URL encode`.

afeena:
How do you test cookie injection? :)
I try some simple cases and run sqlmap, but didn't get any result.
```[17:10:39] [WARNING] Cookie parameter 'sess_uuid' does not seem to be injectable```

rnehra01:
Setting a cookie `id` = `1 and 1 --` is working for me.

afeena:
Oh, I found some problems with setting some cookies in chrome. With Firefox it works well :)

rnehra01:
resolve #150

Issue #100, opened 2017-02-09
Make the unified system for TANNER/SNARE communication


afeena:
We can get differ types of messages from TANNER, so we should check the type and existence of some dict keys to process the page. https://github.com/mushorg/snare/blob/master/snare.py#L187

TANNER should return unified messages to make processing the result of the emulation easier. For example, now payload can be dict or str, so we should make a lot of checks to choose the action. If TANNER returns unified structured messages we will have the possibility to parse it properly. TANNER should have clear, proper api. Also we can return error messages if something goes wrong and make logs (as a option) of these errors on the SNARE side. 

Maybe we can make the types of the messages, for example: 
* Type 1 - it's a common query, no attack, nothing to do; 
* Type 2 - injectable payload (data should be shown on the current page) ; 
* Type 3 -  payload with new page; 
* Type 4 - error. 
An so on.

So, if we get the type 4, we definitely know that we need to get "error" key from the message.

It's a bit dirty description. Any advice and suggestions will be greatly appreciated

glaslos:
Sounds good to me. We could also go similar like HTTP:
- 1 common type, but 101 is index page, 102 is css, 103 is image...
- 2 attack. 201 with inject, 202 with new page...
- 3 error. 301 req error, 302 tanner error...

etc.

Philelis:
I think all this types should also be in the logfiles as severity. Do you agree with me?

afeena:
@Philelis Sure, I completely agree with you, thanks :)

afeena:
#184

PR #153, opened 2017-06-16
Fix Docker Freezing

### LINKED ISSUE #151: Docker freezing
I make small load testing for cmd executor. I use this command 
`cat /dev/urandom`
After that, tanner stops answer any requests.

Docker now is not async. So, this command never stops and never returns the result and tanner waits for the result in blocking mode. We can run docker functions in [executor](https://docs.python.org/3/library/asyncio-eventloop.html#executor)

rnehra01:
I've allowed time of 1sec for execution, if it doesn't get executed within that time, then payload is empty or we can provide something `Process Timeout`
Fix #151

### DIFF
--- a/tanner/emulators/cmd_exec.py
+++ b/tanner/emulators/cmd_exec.py
@@ -2,7 +2,6 @@ import asyncio
 import docker
 import yarl
 import concurrent
-import time
 # TODO : Replace docker with aiodocker
 import logging
 
--- a/tanner/emulators/cmd_exec.py
+++ b/tanner/emulators/cmd_exec.py
@@ -1,6 +1,8 @@
 import asyncio
 import docker
 import yarl
+from concurrent.futures import ThreadPoolExecutor
+import time
 # TODO : Replace docker with aiodocker
 import logging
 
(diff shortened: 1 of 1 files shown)


tanner/emulators/cmd_exec.py
```diff
@@ -51,10 +53,14 @@ def __init__(self):
...
-        execute_result = None
+        execute_result = ''
         try:
             container.start()
-            execute_result = container.exec_run(['sh', '-c', cmd]).decode('utf-8')
+            executor = ThreadPoolExecutor(max_workers=1)
+            cmd_future = executor.submit(container.exec_run, ['sh', '-c', cmd])
+            time.sleep(1)
+            if(cmd_future.done()):
+                execute_result = cmd_future.result().decode('utf-8')
```
afeena: Maybe it is better to use timeout instead of blocking sleep?
https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.result

rnehra01:
And what about payload if it gets blocked, empty string or anything else like `Timeout Error`

afeena:
Maybe we should return error page? Something like 504 error.

rnehra01:
Currently, payload from emulators support `payload_value` and `injected page`. I'm thinking of implementing headers, cookies and response codes in the task proposed (Improve tanner communication) #100

afeena:
So, for the first step we can return empty payload :)

tanner/emulators/cmd_exec.py
```diff
@@ -1,6 +1,8 @@
 import asyncio
 import docker
 import yarl
+import concurrent
+import time
```
afeena: unused import :)

PR #158, opened 2017-06-22
Add docker for LFI

### LINKED ISSUE #1: /proc emulation
Have a look at http://user-mode-linux.sourceforge.net/old/hppfs.html

rnehra01:
This improves LFI emulation as it provides a totally (real) virtual file system.
I guess this fixes #1.


afeena:
I try some lfi paths, and while using NULL byte character ```file=../../../../../etc/passwd%00```, I got this error ```rpc error: code = 2 desc = containerd: container not started```. 
Is it expected behavior? I think with that we give some knowledge about docker for attacker.

rnehra01:
No, it isn't expected. So the container is not handling the '\x00' byte.So the fix can be we can terminate the string with `NULL` which doesn't happen in python like C.

rnehra01:
There is something strange with the last 2nd build, it is using `test_image` for Docker host_image. It is probably because of `test_config` where config is set to that value. Though I've fixed it by explicitly specifying `host_image` in `test_lfi` file.

PR #166, opened 2017-06-29
Add new tanner Api function

rnehra01:
- [x] Get a session info from its `sess-uuid`
- [x] Get all the sessions from a given `ip`, `time-duration`, `user-agent` , `owner-type`, `attack_type`
- [x] Develop access to above methods
- [x] Combined statistics

### DIFF
--- a/tanner/api.py
+++ b/tanner/api.py
@@ -52,7 +52,10 @@ class Api:
                 if sess['sess_uuid'] == sess_uuid:
                     return sess
 
-    async def return_sessions_by_ip(self, redis_client, peer_ip, snare_uuid= None):
+    async def return_sessions(self, redis_client, filters, snare_uuid= None):
+        valid_filters = self.validate_filters(filters)
+        if validate_filters is not dict:
+            return 'Invalid filters'
         query_res = []
         if snare_uuid:
             snare_uuids = [snare_uuid]
@@ -63,21 +66,36 @@ class Api:
         for snare_id in snare_uuids:
             sessions = await self.return_uuid_stats(snare_id, redis_client)
             for sess in sessions:
-                if sess['peer_ip'] == peer_ip:
+                is_matching_sesssion = True
+                if 'user_agent' in valid_filters:
+                    if valid_filters['user_agent'] not in sess['user_agent']:
+                        is_matching_sesssion = False
+                if 'peer_ip' in valid_filters:
+                    if valid_filters['peer_ip'] != sess['peer_ip']:
+                        is_matching_sesssion = False
+                if 'attack_type' in valid_filters:
+                    if valid_filters['attack_type'] not in sess['attack_types']:
+                        is_matching_sesssion = False
+                if 'time_interval' in valid_filters:
+                    if valid_filters['time_interval']['end_time'] > sess['start_time'] or valid_filters['time_interval']['start_time'] < sess['end_time']:
+                        is_matching_sesssion = False
+                if 'owner_type' in valid_filters:
+                    if valid_filters['owner_type'] not in sess['owner_types']:
+                        is_matching_sesssion = False
+
+                if is_matching_sesssion:
                     matching_sessions.append(sess['sess_uuid'])
         return matching_sessions
 
-    async def return_sessions_by_time(self, redis_client, time_interval, snare_uuid= None):
-        query_res = []
-        if snare_uuid:
-            snare_uuids = [snare_uuid]
+    @staticmethod
+    def validate_filters(filters):
+        possible_filters = ['user-agent', 'peer_ip', 'attack_type', 'time_interval', 'owner_type']
+        valid_filters = {}
+        for key, val in filters.items():
+            if key in possible_filters:
+                valid_filters[key] = val
+
+        if 'time_interval' in valid_filters and valid_filters['time_interval'] is dict and 'start_time' in valid_filters['time_interval'] and 'start_time' in valid_filters['time_interval']:
+            return 'Invalid time filter'
         else:
-            snare_uuids = await self.return_stats(redis_client)
-
-        matching_sessions = []
-        for snare_id in snare_uuids:
-            sessions = await self.return_uuid_stats(snare_id, redis_client)
-            for sess in sessions:
-                if time_interval['start_time'] <= sess['start_time'] <= time_interval['end_time'] or time_interval['start_time'] <= sess['end_time'] <= time_interval['end_time']:
-                    matching_sessions.append(sess['sess_uuid'])
-        return matching_sessions
+            return valid_filters
--- a/tanner/api.py
+++ b/tanner/api.py
@@ -13,8 +13,15 @@ class Api:
 
         if query == 'stats' and not params:
             result = await self.return_stats(redis_client)
-        elif query == 'stats' and 'uuid' in params:
+        elif query == 'snare-stats' and 'uuid' in params:
             result = await self.return_uuid_stats(params['uuid'], redis_client, 50)
+        elif query == 'session-stats':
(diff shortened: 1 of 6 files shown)


tanner/api.py
```diff
@@ -13,8 +13,19 @@ def __init__(self):
 
         if query == 'stats' and not params:
             result = await self.return_stats(redis_client)
-        elif query == 'stats' and 'uuid' in params:
+        elif query == 'snare-stats' and 'uuid' in params:
```
afeena: Maybe it is suitable to think about sub-app approach? To make it easier to add api paths. http://aiohttp.readthedocs.io/en/stable/whats_new_1_1.html#sub-applications

rnehra01:
@afeena check this.

tanner/api.py
```diff
@@ -38,3 +72,48 @@ def __init__(self):
...
+                    if filters['peer_ip'] != sess['peer_ip']:
+                        is_matching_sesssion = False
+                if 'attack_type' in filters:
+                    if filters['attack_type'] not in sess['attack_types']:
+                        is_matching_sesssion = False
+                if 'time_interval' in filters:
+                    if filters['time_interval']['end_time'] < sess['start_time'] or filters['time_interval']['start_time'] > sess['end_time']:
+                        is_matching_sesssion = False
+                if 'owner_type' in filters:
+                    if filters['owner_type'] not in sess['owner_types']:
```
afeena: Server error when try to get owners by type. We don't have `owner_types` key in the session info
```
  File "/usr/local/lib/python3.5/dist-packages/aiohttp/web_server.py", line 62, in handle_request
    resp = yield from self._handler(request)
  File "/usr/local/lib/python3.5/dist-packages/aiohttp/web.py", line 270, in _handle
    resp = yield from handler(request)
  File "/usr/local/lib/python3.5/dist-packages/Tanner-0.3.0-py3.5.egg/tanner/api.py", line 50, in handle_sessions
    result = await self.return_sessions(applied_filters)
  File "/usr/local/lib/python3.5/dist-packages/Tanner-0.3.0-py3.5.egg/tanner/api.py", line 146, in return_sessions
    if filters['owner_type'] not in sess['owner_types']:
KeyError: 'owner_types'
```

{'sensor_uuid': 'c6c8a20f-cc3c-490f-a0c1-4919b5e4b746', 'accepted_paths': 1, 'cookies': {'sess_uuid': None}, 'approx_time_between_requests': 0.0, 'start_time': 1489822602.9002578, **'possible_owners': ['user', 'attacker']**, 'requests_in_second': 2.384185791015625e-07, 'end_time': 1489822602.900258, 'peer_port': 53395, 'sess_uuid': '9aa2322035144fdeab7e5006e1019589', 'errors': 0, 'attack_types': [], 'peer_ip': '192.168.56.1', 'paths': [{'path': '/etc/passwd', 'timestamp': 1489822602.9002342, 'response_status': 200}], 'hidden_links': 0, 'user_agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/56.0.2924.87 Safari/537.36'}

afeena:
Good! I like new api sub-app :)

afeena:
Could you please fix tests?

rnehra01:
@afeena done.

afeena:
I've found some problems with filters

tanner/api.py
```diff
@@ -38,3 +100,52 @@ def __init__(self):
...
+                    if filters['attack_type'] not in sess['attack_types']:
+                        is_matching_sesssion = False
+                if 'time_interval' in filters:
+                    if filters['time_interval']['end_time'] < sess['start_time'] or filters['time_interval']['start_time'] > sess['end_time']:
+                        is_matching_sesssion = False
+                if 'possible_owner' in filters:
+                    if filters['possible_owner'] not in sess['possible_owners']:
+                        is_matching_sesssion = False
+
+                if is_matching_sesssion:
```
afeena: default value is True, so for every non-valid filter string it returns snare-uuids (see comment with screenshot)

tanner/api.py
```diff
@@ -38,3 +100,52 @@ def __init__(self):
...
+            snare_uuids = await self.return_snares()
+
+        matching_sessions = []
+        for snare_id in snare_uuids:
+            sessions = await self.return_snare_info(snare_id)
+            for sess in sessions:
+                if sessions == 'Invalid SNARE UUID':
+                    continue
+                is_matching_sesssion = True
+                if 'user-agent' in filters:
```
afeena: I suggest you to use more general approach here
e.g.
for every `key, value` in `filters` return session which `sess[key]==value`
and use `try` `except` for KeyError

tanner/api.py
```diff
@@ -38,3 +100,52 @@ def __init__(self):
...
+            snare_uuids = await self.return_snares()
+
+        matching_sessions = []
+        for snare_id in snare_uuids:
+            sessions = await self.return_snare_info(snare_id)
+            for sess in sessions:
+                if sessions == 'Invalid SNARE UUID':
+                    continue
+                match_count = 0
+                if 'user-agent' in filters:
```
afeena: Can we try more general approach? What if we have more filters? Should we return error when filter if invalid?

rnehra01: `sess[key] == value` doesn't seem to be possible because for e.g in `time_interval` we have to compare two value, in `possible_owner` we have to compare against a `list`

afeena: About time interval - you can split it into two separate filters: `start time`, `end time` (also it gives more freedom to user)
You have 4 operations:
`equal, greater, less, and include (`in`)`
so you can make some function `apply_filter` which get filter_name and operation as parameters and consistently apply filters from the filters list

rnehra01: So I tried making new `method` but in `apply_filter` method you need to use 4 `if` statements which is same as now. And this will make the code more lengthy.

afeena: Just as a quick example, how it could be done with operators
 
```
    def applay_filter(self, sessions, filter_name, value, op):
        ops = {"in":operator.contains,
               "ge":operator.ge,
               "lt":operator.le,
               "eq":operator.eq}
        result = []
        for sess in sessions:
            if sessions == 'Invalid SNARE UUID':
                continue
            try:
                if ops[op](sess[filter_name],value):
                    result.append(sess)
            except KeyError:
                raise

        return result

    async def return_sessions(self, filters):
        avaliable_filters = {"user_agent":"in",
                             "peer_ip":"eq",
                             "attack_type":"in",
                             "start_time":"le",
                             "end_time":"ge",
                             "possible_owners":"in"}

        if 'snare_uuid' in filters:
            snare_uuids = [filters['snare_uuid']]
        else:
            snare_uuids = await self.return_snares()

        for snare_id in snare_uuids:
            sessions = await self.return_snare_info(snare_id)
            result = sessions
            for f_name, value in filters.items():
                try:
                    result = self.applay_filter(result, f_name,value,avaliable_filters[f_name])
                except KeyError:
                    return "invalid filter or parameter"
                else:
                    return [s['sess_uuid'] for s in result]
```
I think we can make a **separate task** for filters design and find the best way to do this task.

But still, I strongly recommend you to use `try-except` in your code :)

rnehra01: The current code is clean, so I guess there is no need for `try/except`. And I too think of adding `filters` somewhat later.

rnehra01:
@afeena take a look. I've changed `sensor` to `snare_uuid`, I hope it won't be a problem.

afeena:
@rnehra01  Nice work! I think in the context of the snare/tanner snare_uuid is more understandable than sensor :)

PR #175, opened 2017-07-17
Make emulator set flexible

rnehra01:
Now emulators can be enabled using config.

### DIFF
--- a/tanner/config.py
+++ b/tanner/config.py
@@ -9,7 +9,9 @@ config_template = {'DATA': {'db_config': '/opt/tanner/db/db_config.json', 'dorks
                             'vdocs': '/opt/tanner/data/vdocs.json'},
                    'TANNER': {'host': '0.0.0.0', 'port': 8090},
                    'REDIS': {'host': 'localhost', 'port': 6379, 'poolsize': 80, 'timeout': 1},
-                   'EMULATORS': {'root_dir': '/opt/tanner'},
+                   'EMULATORS': {'root_dir': '/opt/tanner',
+                                  'emulator_enabled': {'sqli': True, 'rfi': True, 'lfi': True, 'xss': True, 'cmd_exec': True}
+                                },
                    'SQLI': {'type':'SQLITE', 'db_name': 'tanner_db', 'host':'localhost', 'user':'root', 'password':'user_pass'},
                    'DOCKER': {'host_image': 'busybox:latest'},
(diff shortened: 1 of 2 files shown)


rnehra01:
I'll fix test later

afeena:
Emulators are initialized in init function of base emulator before defining the set - I think we should make separate initializing function. The problem with e.g. command exec emulator in it's init function - it calls docker helper which try to connect to docker. If we can't install docker on the machine, it becomes the real problem.  So, there are two ways: initialize emulators after defining the set (as I said before) or change init of CmdExecution

rnehra01:
I guess this will do the trick.

tanner/config.py
```diff
@@ -9,7 +9,9 @@
                             'vdocs': '/opt/tanner/data/vdocs.json'},
                    'TANNER': {'host': '0.0.0.0', 'port': 8090},
                    'REDIS': {'host': 'localhost', 'port': 6379, 'poolsize': 80, 'timeout': 1},
-                   'EMULATORS': {'root_dir': '/opt/tanner'},
+                   'EMULATORS': {'root_dir': '/opt/tanner',
+                                  'emulator_enabled': {'sqli': True, 'rfi': True, 'lfi': True, 'xss': True, 'cmd_exec': True}
```
afeena: Could we store only emulators names as list? And if emulator in the list, it's enabled?

tanner/emulators/base.py
```diff
@@ -45,13 +47,14 @@ def extract_get_data(self, path):
...
-                        attack_params[emulator] = []
-                    attack_params[emulator].append(dict(id= param_id, value= param_value))
+            	if self.emulator_enabled[emulator]:
+	                possible_detection = self.emulators[emulator].scan(param_value) if param_value else None
+	                if possible_detection:
+	                    if detection['order'] < possible_detection['order']:
+	                        detection = possible_detection
+	                    if emulator not in attack_params:
+	                        attack_params[emulator] = []
+	                    attack_params[emulator].append(dict(id= param_id, value= param_value))
```
afeena: space after `=` in params

tanner/config.py
rnehra01: According to me, this is a more appropriate way of enabling or disabling things. If we store as list it is little inconvenient to use, because we have fixed keywords for emulators e.g `cmd_exec` so user has to find out that.

afeena: Ok, I think it's just a matter of taste. :)

PR #176, opened 2017-07-20
Add option for poolsize

rnehra01:
It also takes care too many connections for tannerweb.


rnehra01:
The same issue will need to be resolved for new API server. So I'll add that once #177 will be merged.

rnehra01:
Also, I'll proper web app shutdown later.

PR #177, opened 2017-07-20
New api server

rnehra01:


setup.py
```diff
@@ -9,6 +9,6 @@
       author_email='glastopf@public.honeynet.org',
       url='https://github.com/mushorg/tanner',
       packages=find_packages(exclude=['*.pyc']),
-      scripts=['bin/tanner', 'bin/tannerweb'],
+      scripts=['bin/tanner', 'bin/tannerweb', 'bin/api'],
```
afeena: Maybe it's better to use more specific name than `api`?

rnehra01:
@afeena done.

PR #178, opened 2017-07-23
Update docs

rnehra01:


rnehra01:
@afeena Please check if I've missed something.

afeena:
I think everything ok :)

PR #181, opened 2017-07-27
Improve Web UI

### LINKED ISSUE #179: Web request freezing
When I tried to get sessions, tanner web freezing. I've found that it might be because of the large number of sessions in snare_uuid (42297 in list). When we make request to redis, it takes range from 0 to -1 (end) https://github.com/mushorg/tanner/blob/develop/tanner/api/api.py#L43  I set the count as 10 and everything works well.
I suggest to request less sessions per request.

rnehra01:
- [x] fix #179 Web request freezing
- [x] Limit large no of sessions display on a single page

### DIFF
--- a/tanner/api/api.py
+++ b/tanner/api/api.py
@@ -88,7 +88,7 @@ class Api:
                         return 'Invalid filter : %s' % filter_name
 
                 if match_count == len(filters):
-                    matching_sessions.append(sess['sess_uuid']) 
+                    matching_sessions.append(sess) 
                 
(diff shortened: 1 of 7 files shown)


afeena:
Am I doing something wrong?
xs8xndy 1

rnehra01:
I'm not sure what is the reason on your machine.
image

afeena:
Ok, great. I will test it again after merge :)

Could you please give specific names to the branches next time?

PR #183, opened 2017-08-06
PHP Code Injection Emulator

### LINKED ISSUE #182: PHP Code Injection
An emulator that can emulate PHP code injection vulnerability.(https://www.owasp.org/index.php/Code_Injection)

rnehra01:
FIx #182 
So usually PHP code injection takes place when the input is directly passed to `eval`, `assert` or these type of function which eval'd the code. I choose `eval` to implement it where the input is assigned to a variable using `eval`.


tanner/utils/patterns.py
```diff
@@ -7,7 +7,8 @@
 LFI_FILEPATH = re.compile('((\.\.|\/).*)')
 XSS_ATTACK = re.compile('.*<(.|\n)*?>')
 CMD_ATTACK = re.compile('.*(alias|cat|cd|cp|echo|exec|find|for|grep|ifconfig|ls|man|mkdir|netstat|ping|ps|pwd|uname|wget|touch|while).*')
+PHP_CODE_INJECTION = re.compile('.*(echo|system|print).*')
```
glaslos: This feels almost too loose, maybe find a way to make it more strict?

rnehra01: Now for injection attacker must supply `;` to make a new line of code.

tanner/emulators/php_code_injection.py
```diff
@@ -0,0 +1,36 @@
...
+        code_injection_result = None
+        code = '<?php eval(\'$a = {code}\'); ?>'.format(code=code)
+        try:
+            async with aiohttp.ClientSession(loop=self._loop) as session:
+                async with session.post('http://127.0.0.1:8088/', data=code) as resp:
+                    code_injection_result = await resp.json()
+        except aiohttp.ClientError as client_error:
+            self.logger.error('Error during connection to php sandbox %s', client_error)
+        else:
+            await resp.release()
```
afeena: http://aiohttp.readthedocs.io/en/stable/client_reference.html#aiohttp.ClientResponse.release

and also

https://github.com/aio-libs/aiohttp/blob/44edff8a3a930e4ecc35773a48c300c29b9919ae/aiohttp/client.py#L608

So I think we don't need to call release :)
with statement makes it for us, I suppose

tanner/emulators/php_code_injection.py
```diff
@@ -0,0 +1,36 @@
...
+    def __init__(self, loop=None):
+        self._loop = loop if loop is not None else asyncio.get_event_loop()
+        self.logger = logging.getLogger('tanner.php_code_injecton')
+
+    async def get_injection_result(self, code): 
+        code_injection_result = None
+        code = '<?php eval(\'$a = {code}\'); ?>'.format(code=code)
+        try:
+            async with aiohttp.ClientSession(loop=self._loop) as session:
+                async with session.post('http://127.0.0.1:8088/', data=code) as resp:
```
afeena: We could think about getting this value (phpox address) from config

tanner/emulators/base.py
```diff
@@ -5,21 +5,24 @@
...
             'rfi': rfi.RfiEmulator(base_dir, loop) if self.emulator_enabled['rfi'] else None,
             'lfi': lfi.LfiEmulator() if self.emulator_enabled['lfi'] else None,
             'xss': xss.XssEmulator() if self.emulator_enabled['xss'] else None,
             'sqli': sqli.SqliEmulator(db_name, base_dir) if self.emulator_enabled['sqli'] else None,
-            'cmd_exec': cmd_exec.CmdExecEmulator() if self.emulator_enabled['cmd_exec'] else None
-        }
-        self.get_emulators = ['sqli', 'rfi', 'lfi', 'xss', 'cmd_exec']
-        self.post_emulators = ['sqli', 'rfi', 'lfi', 'xss', 'cmd_exec']
+            'cmd_exec': cmd_exec.CmdExecEmulator() if self.emulator_enabled['cmd_exec'] else None,
+            'php_code_injection': php_code_injection.PHPCodeInjection() if self.emulator_enabled['php_code_injection'] else None
```
afeena: I guess you should pass loop into init function and it  fixes tests

rnehra01:
So I guess nowI can add all the functions present in the [phpox/function.py](https://github.com/mushorg/phpox/blob/master/functions.py)

tanner/emulators/php_code_injection.py
rnehra01: Yeah, we can make the same change in `rfi` also where phpox address is written directly.

afeena:
I've tested a bit and found small problem
If I try something like /index.php?arg=1; phpinfo() error in the phpox occurs because the result code don't have `;` after phpinfo(), but I think it's not a problem :) Just for the information

PR #184, opened 2017-08-12
Snare tanner communication

### LINKED ISSUE #100: Make the unified system for TANNER/SNARE communication

rnehra01:
fix #100
This is how return value from tanner look like : 
```
Case 1 (where you need to return the page normally)
detection = {
		type : 1
        }
Case 2 (inject payload in the page)
detection = {
		type : 2,
payload = {
		page : ‘/vuln.php’,
		value : ‘<script>alert(1)</script>’
		headers : {
				new_header : ‘new_header_value’
			     }
	      }
        }
Case 3 (where input cause some error so return related to the type of error produced 
  e.g if input takes more time than expected then return 50X) 
detection = {
		type : 3,
                payload = {
		                  status_code : 500/504
                                 } 
       }

```

 
To make communication more systematic, return value type from each emulator is fixed i.e `dict(value='', page=True/False)` (if page=False then payload won't be injected on any page but on a new page)

### DIFF
--- a/tanner/emulators/base.py
+++ b/tanner/emulators/base.py
@@ -117,7 +117,7 @@ class BaseHandler:
         if 'payload' not in detection:
             detection['type'] = 1
         elif 'payload' in detection:
-            if 'status_code' not in detection:
+            if 'status_code' not in detection['payload']:
                 detection['type'] = 2
                 if detection['payload']['page']:
                     injectable_page = self.set_injectable_page(session)
--- a/tanner/emulators/base.py
+++ b/tanner/emulators/base.py
@@ -58,11 +58,12 @@ class BaseHandler:
 	                    if emulator not in attack_params:
 	                        attack_params[emulator] = []
 	                    attack_params[emulator].append(dict(id=param_id, value=param_value))
-                    
+
         if detection['name'] in self.emulators:
             emulation_result = await self.emulators[detection['name']].handle(attack_params[detection['name']], session)
-            detection['payload'] = emulation_result
-
+            if emulation_result:
+                detection['payload'] = emulation_result
+                
         return detection
 
     async def handle_post(self, session, data):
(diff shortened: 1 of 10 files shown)


tanner/emulators/base.py
```diff
@@ -113,11 +114,18 @@ def set_injectable_page(session):
...
         
-        if 'payload' in detection and type(detection['payload']) is dict:
-            injectable_page = self.set_injectable_page(session)
-            if injectable_page is None:
-                injectable_page = '/index.html'
-            detection['payload']['page'] = injectable_page
+        if 'payload' not in detection:
+            detection['type'] = 1
+        elif 'payload' in detection:
+            if 'status_code' not in detection:
```
afeena: As I understood from code, `status_code` in detection['payload'], am I right?

tanner/emulators/base.py
```diff
@@ -113,11 +114,18 @@ def set_injectable_page(session):
...
+                detection['type'] = 2
+                if detection['payload']['page']:
+                    injectable_page = self.set_injectable_page(session)
+                    if injectable_page is None:
+                        injectable_page = '/index.html'
+                    detection['payload']['page'] = injectable_page
+            else:
+                detection['type'] = 3
```
afeena: Can we return any status info about error?

tanner/emulators/base.py
rnehra01: Yes, you're right.

tanner/emulators/base.py
rnehra01: Do we need that because usually that will be present in logs where it was present? And even if get it then how will we use it, as suppose we return `500` error to the client then sending any other data won't be needed.

afeena: In case of tanner error, snare should works normally (e.g. return index page). Imagine the situation, when we use snare with default tanner and we don't have access to tanner's log. Something happens, but we don't know what.

rnehra01: Ok, but this thing is when you want to give a particular status_code intentionally to give positive results for the attack, though I'll add that too for normal tanner error.

PR #198, opened 2017-08-24
Some small changes

rnehra01:
- [x] Use phpox address from config
- [x] Return tanner version to snare


rnehra01:
@afeena how do I get tanner version?

afeena:
@rnehra01 You can define `__version__` attribute in package init file.

tanner/emulators/rfi.py
```diff
@@ -76,9 +77,12 @@ def download_file_ftp(self, url):
             return rfi_result
         with open(os.path.join(self.script_dir, file_name), 'br') as script:
             script_data = script.read()
+        phpox_address = 'http://{host}:{port}'.format(host=config.TannerConfig().get('PHPOX', 'host'),
```
afeena: I think `TannerConfig` not a function. Delete `()`

PR #199, opened 2017-09-06
Add hpfeeds logging support

standa4:


tanner/reporting/hpfeeds.py
```diff
@@ -0,0 +1,177 @@
...
+        if type(chaninfo) == str:
+            chaninfo = [chaninfo,]
+        for c in chaninfo:
+            self.s.send(msgpublish(self.ident, c, data))
+
+    def close(self):
+        try: self.s.close()
+        except: logger.warn('Socket exception when closing.')
+
+def new(host=None, port=10000, ident=None, secret=None, reconnect=True, sleepwait=20):
```
afeena: `sleepwait` is not used

standa4:
Hello,
I have fixed the issues you have mentioned.
The code is based on the code from thug honeypot so the issues with indentation and unused variable are from there.

afeena:
Hi @standa4 Thanks for fixing issues :)

tanner/reporting/hpfeeds.py
```diff
@@ -0,0 +1,177 @@
...
+        self.unpacker = FeedUnpack()
+
+        self.connect()
+
+    def connect(self):
+        logger.info('connecting to %s:%s', self.host, self.port)
+        self.s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self.s.settimeout(self.timeout)
+        try: self.s.connect((self.host, self.port))
+        except: raise FeedException('Could not connect to broker.')
```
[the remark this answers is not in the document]
afeena: FeedException is raised, but it is never catched. So if something wrong with the connection, tanner can't start at all.

standa4:
Hello @afeena , 
sorry for the bugs. I have implemented this feature for my need and checked if it works. I wanted to share it with others. Unfortunately I didn't realize to check every possible error state of this code. I will try to fix what I can and I will let you know when I fix it. Thank you and sorry for the problems.

afeena:
@standa4 Thank you so much for sharing and fixing bugs :)

tanner/reporting/hpfeeds.py
```diff
@@ -0,0 +1,224 @@
...
+        except socket.timeout:
+            logger.warn("Timeout while sending - disconnect.")
+            raise Disconnect()
+        except socket.error as e:
+            logger.warn("Socket error: %s", e)
+            raise Disconnect()
+
+    def tryconnect(self):
+        if not self.connected:
+            while True:
```
afeena: I don't think `while True` is a good decision here. In case when the connection is not alive for a long time, tanner can't work because it's totally blocked. There are two ways - exit the tanner with logs and messages or mark the hpfeeds connection unsuccessful, log it and don't use later. We can use first way for now, it will be simpler.

I really appreciate your help and I completely want to merge your pr in master :)

standa4: I know the `while True` is not a best idea and it can cause the problems, but it is from the original hpfeeds lib (https://github.com/rep/hpfeeds/blob/master/lib/hpfeeds.py). I wanted to discuss it with you but I was quite busy last days. I will try to improve it but right now I need to do some other work tasks. Sorry. I will make the fix ASAP.  
Thank you.

standa4:
Hello @afeena,
I have made some changes to the code. I still need to do some testing that I didn't break anything.
I would like to ask what you think about limiting the number of the reconnect attempts. Is it usable or should I think about different solution?
Thank you.

afeena:
Hellp @standa4 
I think it's ok on this stage. :) Let me know when you finish your testing and I'll merge your pr :)

standa4:
Hello,
I made some tests and it seems ok. 
Thank you.

afeena:
@standa4 Thanks for your work :)

PR #205, opened 2017-11-03
Added a very small clarification in the docs

nsmfoo:
Added a very small clarification in the docs, just in case someone fails to read the error messages =)


afeena:
Thanks :)

P.S.  Build was broken because of new version of aiohttp (tests work with aiohttp 2.2, but don't work with 2.3) I'll change required version until I understand the exact problem.

PR #222, opened 2017-12-12
Fix #220

### LINKED ISSUE #220: Config data types
In config there are `True` and `False` values, which are `strings` https://github.com/mushorg/tanner/blob/master/tanner/config.py#L14, 
but for proper if statements we need `bool` values, e.g here https://github.com/mushorg/tanner/blob/master/tanner/emulators/base.py#L14

Currently switching off emulators doesn't affect emulators set.

surajt97:
Added comparison with "True" in if condition

### DIFF
--- a/tanner/config.py
+++ b/tanner/config.py
@@ -11,14 +11,14 @@ config_template = {'DATA': {'db_config': '/opt/tanner/db/db_config.json', 'dorks
                    'API': {'host': '0.0.0.0', 'port': 8092},
                    'REDIS': {'host': 'localhost', 'port': 6379, 'poolsize': 80, 'timeout': 1},
                    'EMULATORS': {'root_dir': '/opt/tanner'},
-                   'EMULATOR_ENABLED': {'sqli': 'True', 'rfi': 'True', 'lfi': 'True', 'xss': 'True', 'cmd_exec': 'True'},
+                   'EMULATOR_ENABLED': {'sqli': True, 'rfi': True, 'lfi': True, 'xss': False, 'cmd_exec': True},
                    'SQLI': {'type':'SQLITE', 'db_name': 'tanner_db', 'host':'localhost', 'user':'root', 'password':'user_pass'},
                    'DOCKER': {'host_image': 'busybox:latest'},
                    'LOGGER': {'log_debug': '/opt/tanner/tanner.log', 'log_err': '/opt/tanner/tanner.err'},
-                   'MONGO': {'enabled': 'False', 'URI': 'mongodb://localhost'},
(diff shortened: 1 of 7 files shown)


afeena:
I'd better change the type of the values in default config dictionary and add the type conversion into the function `get` of the config, according to default values :)

surajt97:
@afeena i have changed the data types to bool in config.py and test_config.py.

afeena:
@surajt97 Nice! But it will works only for template config in code. If there is an external config, tanner will store string values instead of booleans. I suggest you to modify `get` function of `config.py`: extract the type from template config and use appropriate function, e.g.

```
convert_type = type(config_template[section][value])
res = convert_type(Tanner.config.get(section,value))
```
 
It helps us to avoid any problems with types from config :)

rnehra01:
@surajt97 Tests are failing because of this [line](https://github.com/mushorg/tanner/blob/master/tanner/reporting/log_hpfeeds.py#L12). 
Change `self.port = int(config.TannerConfig.get('HPFEEDS', 'PORT'))` to 
`self.port = config.TannerConfig.get('HPFEEDS', 'PORT')`

tanner/config.py
```diff
@@ -11,14 +11,14 @@
...
+                   'EMULATOR_ENABLED': {'sqli': True, 'rfi': True, 'lfi': True, 'xss': True, 'cmd_exec': True},
                    'SQLI': {'type':'SQLITE', 'db_name': 'tanner_db', 'host':'localhost', 'user':'root', 'password':'user_pass'},
                    'DOCKER': {'host_image': 'busybox:latest'},
                    'LOGGER': {'log_debug': '/opt/tanner/tanner.log', 'log_err': '/opt/tanner/tanner.err'},
-                   'MONGO': {'enabled': 'False', 'URI': 'mongodb://localhost'},
-                   'HPFEEDS': {'enabled': 'False', 'HOST': 'localhost', 'PORT': '10000', 'IDENT': '', 'SECRET': '', 'CHANNEL': 'tanner.events'},
-                   'LOCALLOG': {'enabled': 'False', 'PATH': '/tmp/tanner_report.json'},
-                   'CLEANLOG': {'enabled': 'False'}
+                   'MONGO': {'enabled': False, 'URI': 'mongodb://localhost'},
+                   'HPFEEDS': {'enabled': False, 'HOST': 'localhost', 'PORT': '10000', 'IDENT': '', 'SECRET': '', 'CHANNEL': 'tanner.events'},
```
rnehra01: Remove the `'` quote in the `port` field of `HPFEEDS` to make it `INT`

afeena:
@surajt97 Nice work :)

Issue #189, opened 2017-08-18
Session's delete problem


afeena:
In develop branch I've got this error:
```2017-08-17 21:18 ERROR:aiohttp.server:log_exception: Error handling request
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/site-packages/aiohttp/web_protocol.py", line 421, in start
    resp = yield from self._request_handler(request)
  File "/usr/local/lib/python3.5/site-packages/aiohttp/web.py", line 303, in _handle
    resp = yield from handler(request)
  File "/usr/local/lib/python3.5/site-packages/Tanner-0.4.0-py3.5.egg/tanner/server.py", line 51, in handle_event
    data, self.redis_client
  File "/usr/local/lib/python3.5/site-packages/Tanner-0.4.0-py3.5.egg/tanner/session_manager.py", line 18, in add_or_update_session
    await self.delete_old_sessions(redis_client)
  File "/usr/local/lib/python3.5/site-packages/Tanner-0.4.0-py3.5.egg/tanner/session_manager.py", line 75, in delete_old_sessions
    self.sessions.remove(sess)
ValueError: list.remove(x): x not in list```

rnehra01:
Could tell me how to reproduce it?

afeena:
I made a lot of requests from different ip's. I have an idea why this problem may occurs: the delete process runs on request. When tanner tries to delete session, it can be already deleted by another deleting process of the other request (because delay between requests too small)

nsmfoo:
I can just acknowledge that I'm getting the same errors in the tanner.err log. Sadly no more information to add at this time other then that.

afeena:
@nsmfoo thanks, at least I know that it's not a problem of a particular machine :)

viskey98:
@afeena  this issue is fixed by #231

PR #229, opened 2018-03-01
fix for session's delete problem

viskey98:
Fixes issue #189 
Like @afeena 's explanation that this problem could occur because of the simultaneous removal of same process by different requests, we can create a simple fix by checking if  `sess` still exists in the `session` (list), remove if does or ignore if it doesn't.
Please suggest any changes if required :)


viskey98:
Will create new prs from different branches

PR #230, opened 2018-03-01
Add seconds to logger time

mzfr:
As discussed with  @glaslos this will add "seconds" to tanner logger.

Before these changes:
```
2018-03-01 18:04 INFO:asyncio_redis:connection_lost: Redis connection lost
```
After applying these changes:
```
2018-03-01 18:04:27 INFO:asyncio_redis:connection_lost: Redis connection lost
```


glaslos:
thanks for fixing this. Small detail with big impact!

PR #231, opened 2018-03-06
Fix for session's delete problem

viskey98:
Fixes issue #189
Like @afeena 's explanation that this problem could occur because of the simultaneous removal of same process by different requests, we can create a simple fix by checking if sess still exists in the session (list), remove if does or ignore if it doesn't.
Please suggest any changes if required :)


afeena:
Maybe it is better to use try/except here?

viskey98:
@afeena Requested changes have been made, please look into it :)

afeena:
Well done :)

Issue #241, opened 2018-03-31
not able to run tanner web


anjali-shukla:
I have installed docker and everything needed for tanner. Tanner and Tannerapi is running but tannerweb is giving error. The screenshot of error is attached below.
![tannerweb](https://user-images.githubusercontent.com/35517694/38164448-a89ec0ac-3522-11e8-8b7f-a8b31f870925.png)

viskey98:
@anjali-shukla Have you installed the requirements using `pip3 install -r requirements.txt` ? Because if you have a different version of aiohttp installed it might cause some trouble.

afeena:
@anjali-shukla Thanks for the issue. It was a problem with dependency. Now should be fixed by #243 - please, check it :)

anjali-shukla:
yeah , it worked fine. Thank you

PR #243, opened 2018-04-01
aiohttp-jinja2 version

afeena:
set proper version for aiohttp-jinja2. issue #241


xandfury:
@afeena Interestingly `aiohttp` and `yara` are `aiohttp_jinja2` dependencies as well. Can't we just remove them from requirements?

afeena:
@xandfury In my opinion, if we use some library explicitly - it's better to keep it in requirements. It's not a good idea completely rely on some library requirements.

PR #246, opened 2018-04-05
Dbconfig fix

afeena:
Return config setup during install


viskey98:
@afeena can you explain a bit?

afeena:
To setuз database, we need to read configю Where is config?
https://github.com/mushorg/tanner/blob/master/tanner/utils/base_db_helper.py#L15

I guess this line was deleted accidentally

viskey98:
We would get an IOError (file not found) every time we run that line (https://github.com/mushorg/tanner/blob/master/tanner/utils/base_db_helper.py#L15) right?
This PR Is going to fix it right?

PR #249, opened 2018-04-15
Escape seq

afeena:
Format code according python3.6 deprecated behavior
https://docs.python.org/3/whatsnew/3.6.html#deprecated-python-behavior

pycodestyle 2.4.0 changelog:
<...>
Add W605 warning for invalid escape sequences in string literals. PR #676.
<...>


mzfr:
I don't know why pycodestyle didn't raise any errors on escape sequences when I was formating the code. 
But I guess everything will be good now.

PR #252, opened 2018-06-07
Implement "Location" stats 

viskey98:
This PR will add "Location" stats to session stats.
I have re-written `test_session_analyzer.py` and `templates/session.html` accordingly as well.

### DIFF
--- a/tanner/session_analyzer.py
+++ b/tanner/session_analyzer.py
@@ -1,6 +1,7 @@
 import asyncio
 import json
 from urllib.request import urlopen
+from concurrent.futures import ThreadPoolExecutor
 import logging
 import operator
 import socket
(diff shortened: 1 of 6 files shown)


tanner/session_analyzer.py
```diff
@@ -131,3 +137,18 @@ def choose_possible_owner(stats):
         maxval = max(possible_owners.items(), key=operator.itemgetter(1))[1]
         owners = [k for k, v in possible_owners.items() if v == maxval]
         return {'possible_owners': owners}
+
+    @staticmethod
+    def find_location(ip):
```
afeena: why you decided to use synchronous urlopen? why not async function with aiohttp request?

tanner/tests/test_session_analyzer.py
```diff
@@ -57,3 +57,28 @@ def test_create_stats(self):
         redis_mock.lpush = push_list
         stats = self.loop.run_until_complete(self.handler.create_stats(self.session, redis_mock))
         self.assertEqual(stats['possible_owners'], ['attacker'])
+
+    def test_ip_locator(self):
```
afeena: test names should reflect the function name and status, you do not have object ip_locator, you have find_location

tanner/session_analyzer.py
afeena: And the most important: "__deprecation_message__":"This API endpoint is deprecated and will stop working on July 1st, 2018. For more information please visit: https://github.com/apilayer/freegeoip#readme"

PR #256, opened 2018-06-18
Improve choose_possible_owner function

viskey98:
This PR will improve tanner analyzer based on a heuristic approach. This PR also modifies the tests for the same.

### DIFF
--- a/tanner/session_analyzer.py
+++ b/tanner/session_analyzer.py
@@ -1,7 +1,6 @@
 import asyncio
 import json
 import logging
-import operator
 import socket
 from geoip2.database import Reader
 import geoip2
@@ -43,7 +42,7 @@ class SessionAnalyzer:
 
     async def create_stats(self, session, redis_client):
         sess_duration = session['end_time'] - session['start_time']
-        rps = sess_duration / session['count']
+        rps = session['count'] / sess_duration
         location_info = await self._loop.run_in_executor(
             None, self.find_location, session['peer']['ip']
         )
@@ -69,7 +68,7 @@ class SessionAnalyzer:
             cookies=session['cookies']
         )
 
-        owner = self.choose_possible_owner(stats)
+        owner = await self.choose_possible_owner(stats)
         stats.update(owner)
         return stats
 
@@ -96,13 +95,12 @@ class SessionAnalyzer:
                 attack_types.append(path['attack_type'])
         return tbr_average, errors, hidden_links, attack_types
 
-    @staticmethod
-    def choose_possible_owner(stats):
+    async def choose_possible_owner(self, stats):
         possible_owners = dict(
-            user=0,
-            tool=0,
-            crawler=0,
-            attacker=0
+            user=0.0,
+            tool=0.0,
+            crawler=0.0,
+            attacker=0.0
         )
         attacks = {'rfi', 'sqli', 'lfi', 'xss'}
         bots_owner = ['Mozilla/5.0 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)',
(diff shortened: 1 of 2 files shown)


tanner/session_analyzer.py
```diff
@@ -114,26 +112,14 @@ def choose_possible_owner(stats):
...
-            possible_owners['crawler'] += 1
-        else:
-            possible_owners['user'] += 1
-            possible_owners['attacker'] += 1
-        if stats['hidden_links'] > 0:
-            possible_owners['crawler'] += 1
-            possible_owners['attacker'] += 1
-        if set(stats['attack_types']).intersection(attacks):
-            possible_owners['attacker'] += 1
+        possible_owners['crawler'], possible_owners['tool'] = await self.detect_crawler(stats, bots_owner)
```
afeena: since you use bots_owner only in one function, it's better to move it into it instead of pass it as a parameter

tanner/session_analyzer.py
```diff
@@ -96,13 +95,12 @@ def __init__(self, loop=None):
                 attack_types.append(path['attack_type'])
         return tbr_average, errors, hidden_links, attack_types
 
-    @staticmethod
-    def choose_possible_owner(stats):
+    async def choose_possible_owner(self, stats):
         possible_owners = dict(
```
afeena: dictionary with the same values can be defined as a dict comprehensions

tanner/session_analyzer.py
```diff
@@ -150,3 +136,36 @@ def find_location(ip):
...
+                return (1.0, 0.0)
+        if stats['requests_in_second'] > 10:
+            if stats['user_agent'] in bots_owner:
+                return (0.85, 0.15)
+            return (0.5, 0.85)
+        if stats['user_agent'] in bots_owner:
+            hostname, _, _ = await self._loop.run_in_executor(
+                None, socket.gethostbyaddr, stats['peer_ip']
+            )
+            if 'search.msn.com' or 'googlebot.com' in hostname:
```
afeena: can we extend this list?

tanner/session_analyzer.py
```diff
@@ -150,3 +136,36 @@ def find_location(ip):
         except geoip2.errors.AddressNotFoundError:
             info = "NA"  # When IP doesn't exist in the db, set info as "NA - Not Available"
         return info
+
+    async def detect_crawler(self, stats, bots_owner):
+        for _, path in enumerate(stats['paths']):
```
afeena: why you need to enumerate if you skip index value?

tanner/session_analyzer.py
```diff
@@ -114,26 +112,14 @@ def choose_possible_owner(stats):
...
-        if set(stats['attack_types']).intersection(attacks):
-            possible_owners['attacker'] += 1
+        possible_owners['crawler'], possible_owners['tool'] = await self.detect_crawler(stats, bots_owner)
+        possible_owners['attacker'] = await self.detect_attacker(stats, bots_owner, attacks)
+
+        maxcf = max([possible_owners['crawler'], possible_owners['attacker'], possible_owners['tool']])
 
-        maxval = max(possible_owners.items(), key=operator.itemgetter(1))[1]
-        owners = [k for k, v in possible_owners.items() if v == maxval]
+        possible_owners['user'] = round(1 - maxcf, 2)
```
afeena: could you please explain the purpose of round? What is wrong with simple subtraction?

tanner/session_analyzer.py
viskey98: @afeena bots_owner is used in detect_attacker function also (please check line no.148). That's why I passed it as parameter for both the functions.

tanner/session_analyzer.py
viskey98: I think we can create a new PR, in which we can add many new bot owners and host names, also remove any outdated bots. What do you think about this?

tanner/session_analyzer.py
viskey98: This is because, In other owner cases we assign the cf values ourselves with 2 decimal points. Where as in this case, we do a floating point subtraction, which can (or might) lead to several decimal points, sometimes it might even contain dummy values in the end. 
I found this in a case where i expected cf-user to be `0.15`, but the value came out to be `0.15000000000000002`. This is why I used `round` function to round off the values to 2 decimal points, to avoid any errors or confusions.

tanner/session_analyzer.py
afeena: My bad :)

PR #263, opened 2018-07-17
Tanner develop fixes phase #2

viskey98:
This will fix the following:
- JSON error due to different python version.
- NoneType user-agent error.

### DIFF
--- a/tanner/api/api.py
+++ b/tanner/api/api.py
@@ -51,7 +51,7 @@ class Api:
             if not query_res:
                 return 'Invalid SNARE UUID'
             for (i, val) in enumerate(query_res):
-                query_res[i] = json.loads(val.decode("utf-8"))
+                query_res[i] = json.loads(val)
         return query_res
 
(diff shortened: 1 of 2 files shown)


tanner/api/api.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, redis_client):
             if not query_res:
                 return 'Invalid SNARE UUID'
             for (i, val) in enumerate(query_res):
-                query_res[i] = json.loads(val)
+                query_res[i] = json.loads(val.decode("utf-8"))
```
afeena: I'd better put encoding things into redis query

viskey98: `zrevrangebyscore` redis query doesn't have encoding parameter. So except for this case, I have added encoding things in the redis query itself. Please look into it.

afeena: Actually it has this parameter (maybe they forgot to add it into the docs) :)

viskey98:
@afeena That redis encoding is fixed. Please look into it now :)

PR #264, opened 2018-07-17
Cmd pattern

afeena:
@viskey98 observer that path `/news?color=https://evil.com/shell.php` is matched as cmd_attack.

The problem is in cmd attack patter - it matches everything with specific command
e.g
`https://` is matched with cmd attack because of `ps`

### DIFF
--- a/tanner/utils/patterns.py
+++ b/tanner/utils/patterns.py
@@ -6,7 +6,7 @@ LFI_ATTACK = re.compile(r'.*(/\.\.)*(home|proc|usr|etc)/.*')
 LFI_FILEPATH = re.compile(r'((\.\.|/).*)')
 XSS_ATTACK = re.compile(r'.*<(.|\n)*?>')
 CMD_ATTACK = re.compile(
-    r'.*(alias|cat|cd|cp|echo|exec|find|for|grep|ifconfig|ls|man|mkdir|netstat|ping|ps|pwd|uname|wget|touch|while).*')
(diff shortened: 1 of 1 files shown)


viskey98:
@afeena can you explain a bit this fix? :)

afeena:
@viskey98  The idea is that the command can't be a part of the word (like in the case of https)
Also we can combine a command via e.g. `; | &`, so I just excluded the most obvious cases, e.g. if someone wants to have cat.jpg it will not be a cmd attack :)

tanner/utils/patterns.py
```diff
@@ -6,7 +6,9 @@
 LFI_FILEPATH = re.compile(r'((\.\.|/).*)')
 XSS_ATTACK = re.compile(r'.*<(.|\n)*?>')
 CMD_ATTACK = re.compile(
-    r'.*(alias|cat|cd|cp|echo|exec|find|for|grep|ifconfig|ls|man|mkdir|netstat|ping|ps|pwd|uname|wget|touch|while).*')
+    r'.*[^A-z:.]'
+    r'(alias|cat|cd|cp|echo|exec|find|for|grep|ifconfig|ls|man|mkdir|netstat|ping|ps|pwd|uname|wget|touch|while)'
+    r'[^A-z:.]?.*')
```
viskey98: Can you explain the use of `?` in this regexp. I think it shouldn't be present. Because we definitely don't want the presence of the set `[^A-z:.]` in the end of the matched keyword. `?` will make it a '0-1' case.
False-negative case for above exp:
`/news?color=https://www.evil.com/catalonia.php`
- Expected: No-Match
- Result: Match-found

afeena: You are right :) My problem was in the end of the string, because command can be performed without  arguments. Now it should work

PR #277, opened 2018-10-15
Applied logging exceptions (Fix #238)

### LINKED ISSUE #238: Logging exceptions better
Instead of using `logging.error` use `logging.exception`. [`logger.exception`](http://docs.python.org/library/logging.html#logging.Logger.exception) will output a stack trace alongside the error message. With `logging.error` only the formatted error message would display and the traceback would be thrown away.

More info: https://stackoverflow.com/questions/5191830/how-do-i-log-a-python-error-with-debug-information?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

CyberDrudge:
This fixes #238 
Instead of using `logging.error`, `logging.exception` has been used. With `logging.error` only the formatted error message would display and the traceback would be thrown away, whereas with `logging.exception` exception info is added to the logging message.

### DIFF
--- a/tanner/api/api.py
+++ b/tanner/api/api.py
@@ -14,7 +14,7 @@ class Api:
         try:
             query_res = await self.redis_client.smembers('snare_ids', encoding='utf-8')
         except aioredis.ProtocolError as connection_error:
-            self.logger.error('Can not connect to redis %s', connection_error)
+            self.logger.exception('Can not connect to redis %s', connection_error)
         return list(query_res)
 
     async def return_snare_stats(self, snare_uuid):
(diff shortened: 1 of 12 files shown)


glaslos:
LGTM, the test is failing on a style check where you violate the 120 character line length restriction. @afeena is this a restriction you want to keep enforcing?

afeena:
@glaslos Actually we can change it, but it's even more now than pep8 says. What is your opinion? :)

afeena:
@xandfury No special reasons, feel free to improve :)

PR #281, opened 2018-11-27
fix: Change references to elisabeth to mimesis

leftshift:
When the change from elisabeth to mimesis were made, the references to elisabeth were't actually changed.

Replaces all occurances of `elisabeth` with the according calls to `mimesis`.


afeena:
@leftshift Nice, thanks!

PR #282, opened 2018-11-27
Docker fixes

leftshift:
The docker files and `docker-compose.yml` had various little problems:
* The config in `docker/tanner/dist/config.py` wasn't up-to-date, so tanner didn't actually start.
Overriding the python config file is probably not the best way to do this anyways, but this fixes it for now
* `phpox` didn't work due to a missing package and different binary names
* various directories needed to be writable

Also, I backported [`phpox/#5`](https://github.com/mushorg/phpox/pull/5). Again, this is required because `sandbox.py` is overwritten to change config values, which probably isn't the nicest way to do this


afeena:
Thanks for the contribution! (and sorry for long waiting)

PR #289, opened 2019-01-08
Increases Redis-Server somaxconn default TCP backlog settings to 511 instead of 128

DanielSpring:
Update docker-compose.yml


docker/docker-compose.yml
```diff
@@ -9,6 +9,9 @@ services:
   tanner_redis:
     build: ./redis
     container_name: tanner_redis
+    privileged: true
```
glaslos: is this required?

DanielSpring: privileged: true does not seem to be required.

glaslos: then please remove 👍

glaslos:
Closing this now. Please submit a more isolated PR next time. You are changing multiple things at the same time some of which should not in this PR (Dockerfile, config changes...). I'm happy to help with a new PR

PR #291, opened 2019-03-17
added expiry_time 

amanjiofficial:
expiry_time is added


tanner/config.py
```diff
@@ -15,7 +15,7 @@
                    'WEB': {'host': '0.0.0.0', 'port': 8091},
                    'API': {'host': '0.0.0.0', 'port': 8092},
                    'PHPOX': {'host': '0.0.0.0', 'port': 8088},
-                   'REDIS': {'host': 'localhost', 'port': 6379, 'poolsize': 80, 'timeout': 1},
+                   'REDIS': {'host': 'localhost', 'port': 6379, 'poolsize': 80, 'timeout': 1, 'expiry_time': True},
```
afeena: I highly recommend to make it `false` by default

tanner/redis_client.py
```diff
@@ -15,6 +15,7 @@ class RedisClient:
         try:
             host = TannerConfig.get('REDIS', 'host')
             port = TannerConfig.get('REDIS', 'port')
+            expiry_time = TannerConfig.get('REDIS', 'expiry_time')
```
afeena: For now this variable is not used at all. If you want to add this variable to config, please add the functionality as well.

amanjiofficial: It will be used to auto-delete sessions in redis server so as to make free space

afeena: I understand the purpose of this variable :) Do you want to implement this functionality? If yes, please make it as one PR.

amanjiofficial: Yes I want to implement this functionality. Okay, I will make it as another PR. Is there any better way to implement this functionality ?Suggestions pls

amanjiofficial:
Is this correct way to implement this functionality @afeena @rnehra01

afeena:
Hi @amanjiofficial ,
The idea is to create separate worker to delete the old sessions from Redis. Now you just do not store them. And you can't use boolean value as  `parser` argument in `create_redis_pool`
https://aioredis.readthedocs.io/en/v1.2.0/api_reference.html#aioredis.create_redis_pool

You can try to look at this function: https://redis.io/commands/expire

PR #295, opened 2019-03-23
XSS tests

rjt-gupta:
Increased XSS emulator code coverage to 100%.


tanner/tests/test_xss_emulator.py
```diff
@@ -11,6 +11,12 @@ def setUp(self):
         asyncio.set_event_loop(None)
         self.handler = xss.XssEmulator()
 
+    def test_scan(self):
```
afeena: please add also test with negative result

rjt-gupta:
Added the test

PR #296, opened 2019-03-23
lfi emulator coverage

rjt-gupta:
LFI Emulator coverage increased to 100%.


tanner/tests/test_lfi_emulator.py
```diff
@@ -10,6 +10,12 @@ def setUp(self):
         self.handler = lfi.LfiEmulator()
         self.handler.helper.host_image = 'busybox:latest'
 
+    def test_scan(self):
```
afeena: Can you please also add the case with negative result?

rjt-gupta: Okay, will do.

rjt-gupta:
Negative Test added.

PR #299, opened 2019-03-24
SQLI emulator tests

rjt-gupta:
Changes this PR proposes - 
 - Added missing tests for sqli emulator, overall coverage increased.
 - Tests of `sqli.py` and `sqlite.py` are clashing because of making changes in same file `test_db`. So made some changes in `sqlite.py`.  
Is there better way to do this?

### DIFF
--- a/tanner/tests/test_sqli.py
+++ b/tanner/tests/test_sqli.py
@@ -20,6 +20,8 @@ class SqliTest(unittest.TestCase):
         }
         self.handler = sqli.SqliEmulator('test_db', '/tmp/')
         self.filename = '/tmp/db/test_db'
+        os.makedirs(os.path.dirname(self.filename), exist_ok=True)
+        open('/tmp/db/test_db', 'a').close()
         self.handler.query_map = query_map
         self.sess = mock.Mock()
         self.sess.sess_uuid.hex = 'd877339ec415484987b279469167af3d'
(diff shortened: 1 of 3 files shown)


afeena:
@rjt-gupta You can create handler in each test case, so you can separate sqlite3 and mysql

rjt-gupta:
Okayy, I understand. 
We are using `test_db` as like a default in the other part of project, it is better to keep it consistent.

rjt-gupta:
> 
> 
> @rjt-gupta You can create handler in each test case, so you can separate sqlite3 and mysql

Actually, `test_sqlite.py` and `test_sqli.py` both uses handler of `SQLITEEmulator` and make changes to `test_db` which is causing the problem.

afeena:
@rjt-gupta you can remove test db after tests are complete in `tearDown()`
 https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDown

rjt-gupta:
Perfect.

afeena:
@rjt-gupta please fix test error before I will merge

tanner/tests/test_sqli.py
```diff
@@ -18,7 +19,22 @@ def setUp(self):
             'comments': [{'name': 'comment', 'type': 'text'}]
         }
         self.handler = sqli.SqliEmulator('test_db', '/tmp/')
+        self.filename = '/tmp/db/test_db'
```
afeena: create the file before use

tanner/tests/test_sqli.py
```diff
@@ -18,7 +19,22 @@ def setUp(self):
...
         self.handler.query_map = query_map
+        self.sess = mock.Mock()
+        self.sess.sess_uuid.hex = 'd877339ec415484987b279469167af3d'
+
+    def test_scan(self):
```
afeena: please don't forget negative result test ;)

tanner/tests/test_sqli.py
rjt-gupta: Oops!

rjt-gupta:
updated changes.

PR #305, opened 2019-04-05
add option to change default directory

teeann:


shin2903:
All checks have failed because Python 3.5 doesn't support f-string.

bin/tanner
```diff
@@ -17,7 +17,9 @@ def main():
     """)
     parser = argparse.ArgumentParser()
     parser.add_argument("--config", help="tanner config")
+    parser.add_argument("--base-dir", help="tanner base directory", default='/opt/tanner')
```
rnehra01: Newline

rnehra01:
@afeena Should we create a default dir that doesn't require sudo privileges?

bin/tanner
rnehra01: Add an empty line after this just like we had before.

rnehra01:
@afeena You are talking about [this](https://github.com/mushorg/tanner/blob/master/setup.py#L13), I guess. I'm not sure how to to do that. @shin2903 maybe can you take a look?

bin/tanner
shin2903: done :)

afeena:
For now I suggest to use `.local/share/tanner` or `.tanner` in home directory since we use it only for storing `data` by default :)

glaslos:
@gaochua did you see the suggestion on the default directory?

gaochua:
@glaslos Yes I am working to change the default directory in `setup.py`. This may take a while, as I'm quite busy at the moment.

afeena:
please open new PR if needed

Issue #306, opened 2019-04-07
Dependency version


rjt-gupta:
Snare uses aiohttp_jinja2 latest 1.1.0 while tanner uses version 0.14.0. As version 1.1.0 requires aiohttp 3.2 can we do the upgrade?

rjt-gupta:
Yes, everything looks good to me.

afeena:
@rjt-gupta in this case, just go ahead with pr :)

PR #311, opened 2019-04-19
Upgrade version

rjt-gupta:
refs - #306


rjt-gupta:
Is there anything else needs to be done here?

PR #313, opened 2019-05-13
Base emulator tests

rjt-gupta:


tanner/tests/test_base.py
```diff
@@ -124,3 +124,26 @@ def test_set_injectable_page(self):
             sess = session.Session(None)
         injectable_page = self.handler.set_injectable_page(sess)
         self.assertEqual(injectable_page, '/python.html')
+
+    def test_emulate(self):
+        data = dict(method='GET', path='/path.html?file=/etc/passwd',
+                    cookies={'sess_uuid': '9f82e5d0e6b64047bba996222d45e72c'})
+
+        async def mock_lfi_handle(attack_value, session):
```
rnehra01: To test a certain method, you can mock other methods as they will have their tests separately  so no need for all this just mock `handle_get`

rnehra01: You need to check for different results from `handle_get` will you get correct `type`, will `set_injectable_page` be called with correct params?

rjt-gupta: Perfect!

rjt-gupta:
@rnehra01 Updated changes.

tanner/tests/test_base.py
```diff
@@ -124,3 +124,53 @@ def test_set_injectable_page(self):
...
+        self.handler.set_injectable_page = mock.create_autospec(self.handler.set_injectable_page)
+        data = dict(method='GET', path='/index.html',
+                    cookies={'sess_uuid': '9f82e5d0e6b64047bba996222d45e72c'})
+
+        async def mock_handle_get(session, data):
+            return {'name': 'index', 'order': 2}
+
+        self.handler.handle_get = mock_handle_get
+        detection = self.loop.run_until_complete(self.handler.emulate(data, self.session))
+        assert_detection = {'name': 'index', 'order': 2, 'type': 1, 'version': '0.6.0'}
```
rnehra01: Order for `index` should be 1, I guess. Also, don't hardcode the version otherwise tests will fail once we upgrade the version.

tanner/tests/test_base.py
```diff
@@ -124,3 +124,53 @@ def test_set_injectable_page(self):
...
+        self.handler.handle_get = mock_handle_get
+        self.handler.set_injectable_page = mock_injectable_path
+
+        detection = self.loop.run_until_complete(self.handler.emulate(data, self.session))
+        assert_detection = {'name': 'lfi', 'order': 2, 'payload': {'page': '/path.html'},
+                            'type': 2, 'version': '0.6.0'}
+        self.assertEqual(detection, assert_detection)
+
+        self.handler.set_injectable_page = mock.create_autospec(self.handler.set_injectable_page)
+        self.loop.run_until_complete(self.handler.emulate(data, self.session))
```
rnehra01: Why are you calling it again, just use the result from the previous one and set the mock method for the previous call.

tanner/tests/test_base.py
```diff
@@ -124,3 +124,53 @@ def test_set_injectable_page(self):
...
+        self.handler.set_injectable_page.assert_called_with(self.session)
```
rnehra01: I don't think parameters are important for this method call, so if possible you can avoid that and just use `assert_called_with`.

tanner/tests/test_base.py
```diff
@@ -124,3 +124,53 @@ def test_set_injectable_page(self):
...
+
+    def test_emulate_type_3(self):
+        self.handler.set_injectable_page = mock.create_autospec(self.handler.set_injectable_page)
+        data = dict(method='GET', path='/index.html?file=/etc/passwd',
+                    cookies={'sess_uuid': '9f82e5d0e6b64047bba996222d45e72c'})
+
+        async def mock_handle_get(session, data):
+            return {'name': 'lfi', 'order': 2, 'payload': {'status_code': 200}}
```
rnehra01: Use an example which is currently possible. https://github.com/mushorg/tanner/blob/master/tanner/emulators/php_code_injection.py#L39

tanner/tests/test_base.py
rjt-gupta: Yes this is the problem I was discussing, it is really hard to mock a function and simultaneously assert if it is called because `MagicMock`. Can we merge #314 first?

tanner/tests/test_base.py
rjt-gupta: Got it!

tanner/tests/test_base.py
rnehra01: Done.

rjt-gupta: Cool!

rjt-gupta:
The `set_injectable_page` is static so it can't be mocked using `Asyncmock`. Also, can not perform a `assert_called_once_with` function on a vanilla function: so we need to wrap it with `mock.create_autospec`.
I think it can only be done if I call the loop for them separately.

rnehra01:
> The `set_injectable_page` is static so it can't be mocked using `Asyncmock`. So, to perform `assert_called_with` function on a vanilla function we need to wrap it with `mock.create_autospec`.
> I think it can only be done if I call the loop for them separately. What do you think?

[Here](https://docs.python.org/3/library/unittest.mock.html#quick-guide), they are using `assert_called_with` for static method, did you try this?

rjt-gupta:
Updated! 
I Should have catch that before.

rnehra01:
> Updated!
> 
> It seems `assert_called()` is new in version `3.6`. Should I update the travis file(current - `3.5.6`) or change the tests?

No, keep the python version same.

rjt-gupta:
> 
> 
> > Updated!
> > It seems `assert_called()` is new in version `3.6`. Should I update the travis file(current - `3.5.6`) or change the tests?
> 
> No, keep the python version same.

Cool. changed the tests.

tanner/tests/test_base.py
```diff
@@ -138,7 +137,6 @@ def test_emulate_type_1(self):
         detection = self.loop.run_until_complete(self.handler.emulate(data, self.session))
         assert_detection = {'name': 'index', 'order': 1, 'type': 1, 'version': tanner_version}
         self.assertEqual(detection, assert_detection)
-        self.handler.set_injectable_page.assert_not_called()
```
rnehra01: Why did you remove this?

rjt-gupta: `assert_not_called()` is not available in version `3.5`. It was working in `3.6.8`

rnehra01: Can you try [this](https://stackoverflow.com/questions/12187122/assert-a-function-method-was-not-called-using-mock)?

rnehra01: Also, if this works then add a TODO to update when the python version is upgraded.

rnehra01: Also, I think you lack the habit of googling before raising concerns. For this, I get the above link as the second result by just typing `python mock assert not called `

rjt-gupta: Honestly, its not liked that. It just got very messy for me bcoz of working on too many files at once. But, I will be more careful from now.

PR #315, opened 2019-05-16
MySQL DB Helper Tests

rjt-gupta:
Some pointers - 
- `insert_dummy_data` method will be added in `test_mysqli`.
- multiple files has been generated and they are deleted simultaneously as deleting in `tearDown` phase would not work properly.

### DIFF
--- a/tanner/tests/test_mysql_db_helper.py
+++ b/tanner/tests/test_mysql_db_helper.py
@@ -95,7 +95,7 @@ class TestMySQLDBHelper(unittest.TestCase):
             self.result = await self.cursor.fetchall()
 
         self.loop.run_until_complete(test())
-        assert self.result == self.expected_result
+        self.assertEqual(self.result, self.expected_result)
         assert self.handler.insert_dummy_data.called
 
     @mock.patch('tanner.config.TannerConfig.get', side_effect=mock_config)
@@ -103,12 +103,12 @@ class TestMySQLDBHelper(unittest.TestCase):
         self.expected_result = 1
         self.expected_outs = b''
 
-        dump1 = 'mysqldump --skip-comments --skip-extended-insert -h {host} -u {user} -p{password} ' \
+        dump1 = 'mysqldump --compact --skip-extended-insert -h {host} -u {user} -p{password} ' \
                 'test_db>/tmp/db/file1.sql'
         dump1 = dump1.format(host=TannerConfig.get('SQLI', 'host'),
                              user=TannerConfig.get('SQLI', 'user'),
                              password=TannerConfig.get('SQLI', 'password'))
-        dump2 = "mysqldump --skip-comments --skip-extended-insert -h {host} -u {user} -p{password} " \
+        dump2 = "mysqldump --compact --skip-extended-insert -h {host} -u {user} -p{password} " \
                 "attacker_db>/tmp/db/file2.sql"
         dump2 = dump2.format(host=TannerConfig.get('SQLI', 'host'),
                              user=TannerConfig.get('SQLI', 'user'),
@@ -133,11 +133,8 @@ class TestMySQLDBHelper(unittest.TestCase):
             dump_db_2 = subprocess.Popen(dump2, stdin=subprocess.PIPE, stdout=subprocess.PIPE, shell=True)
             diff_db = subprocess.Popen(diff_db, stdin=subprocess.PIPE, stdout=subprocess.PIPE, shell=True)
             self.outs, errs = diff_db.communicate(timeout=15)
-            dump_db_1.stdout.close()
             dump_db_1.wait()
-            dump_db_2.stdout.close()
             dump_db_2.wait()
-            diff_db.stdout.close()
             diff_db.wait()
 
         except subprocess.CalledProcessError:
--- a/tanner/tests/test_mysql_db_helper.py
+++ b/tanner/tests/test_mysql_db_helper.py
@@ -84,13 +84,18 @@ class TestMySQLDBHelper(unittest.TestCase):
         def mock_read_config():
             return config
 
+        self.expected_result = (('ID', 'int(11)', 'NO', 'PRI', None, ''), ('USERNAME', 'text', 'YES', '', None, ''))
         self.handler.read_config = mock_read_config
         self.handler.insert_dummy_data = AsyncMock()
 
         async def test():
             await self.handler.setup_db_from_config()
+            await self.cursor.execute('USE test_db')
+            await self.cursor.execute('DESCRIBE TEST')
+            self.result = await self.cursor.fetchall()
 
         self.loop.run_until_complete(test())
+        assert self.result == self.expected_result
         assert self.handler.insert_dummy_data.called
 
     @mock.patch('tanner.config.TannerConfig.get', side_effect=mock_config)
@@ -128,8 +133,11 @@ class TestMySQLDBHelper(unittest.TestCase):
             dump_db_2 = subprocess.Popen(dump2, stdin=subprocess.PIPE, stdout=subprocess.PIPE, shell=True)
             diff_db = subprocess.Popen(diff_db, stdin=subprocess.PIPE, stdout=subprocess.PIPE, shell=True)
             self.outs, errs = diff_db.communicate(timeout=15)
+            dump_db_1.stdout.close()
             dump_db_1.wait()
+            dump_db_2.stdout.close()
             dump_db_2.wait()
+            diff_db.stdout.close()
             diff_db.wait()
 
         except subprocess.CalledProcessError:
--- a/tanner/tests/test_mysql_db_helper.py
+++ b/tanner/tests/test_mysql_db_helper.py
@@ -27,22 +27,42 @@ class TestMySQLDBHelper(unittest.TestCase):
         self.returned_result = None
         self.query_map = None
         self.handler = MySQLDBHelper()
+        self.conn = None
+        self.cursor = None
+
+        with mock.patch('tanner.config.TannerConfig.get', side_effect=mock_config) as m:
+            async def connect():
+                self.conn = await self.handler.connect_to_db()
+                self.cursor = await self.conn.cursor()
+
+                # Delete DB if exists
+                self.returned_result = await self.handler.check_db_exists(self.db_name)
+                if self.returned_result == 1:
+                    await self.handler.delete_db(self.db_name)
+
+            self.loop.run_until_complete(connect())
 
     @mock.patch('tanner.config.TannerConfig.get', side_effect=mock_config)
     def test_check_db_exists(self, m):
+        self.expected_result = 1
+
+        async def test():
+            await self.cursor.execute('CREATE DATABASE test_db')
+            await self.conn.commit()
+            self.returned_result = await self.handler.check_db_exists(self.db_name)
+
+        self.loop.run_until_complete(test())
+        self.assertEqual(self.expected_result, self.returned_result)
+
+    @mock.patch('tanner.config.TannerConfig.get', side_effect=mock_config)
+    def test_check_no_db_exists(self, m):
         self.expected_result = 0
 
         async def test():
             self.returned_result = await self.handler.check_db_exists(self.db_name)
 
         self.loop.run_until_complete(test())
-
-        if self.returned_result == 1:
-            async def delete():
-                await self.handler.delete_db(self.db_name)
-            self.loop.run_until_complete(delete())
-        else:
-            self.assertEqual(self.expected_result, self.returned_result)
+        self.assertEqual(self.expected_result, self.returned_result)
 
     @mock.patch('tanner.config.TannerConfig.get', side_effect=mock_config)
     def test_create_query_map(self, m):
--- a/tanner/tests/test_mysql_db_helper.py
+++ b/tanner/tests/test_mysql_db_helper.py
@@ -26,6 +26,7 @@ class TestMySQLDBHelper(unittest.TestCase):
         self.db_name = 'test_db'
         self.expected_result = None
         self.returned_result = None
+        self.result = None
         self.query_map = []
         self.handler = MySQLDBHelper()
         self.conn = None
@@ -89,6 +90,16 @@ class TestMySQLDBHelper(unittest.TestCase):
 
         self.loop.run_until_complete(test())
         assert self.handler.insert_dummy_data.called
+
+    @mock.patch('tanner.config.TannerConfig.get', side_effect=mock_config)
+    def test_copy_db(self, m):
+
+        async def test():
+            self.returned_result = await self.handler.copy_db(self.db_name, "attacker_db")
+            self.result = await self.handler.check_db_exists("attacker_db")
+
+        self.loop.run_until_complete(test())
+        self.assertEqual(self.result, 1)
 
     @mock.patch('tanner.config.TannerConfig.get', side_effect=mock_config)
     def test_create_query_map(self, m):
--- a/tanner/tests/test_mysql_db_helper.py
+++ b/tanner/tests/test_mysql_db_helper.py
@@ -7,12 +7,9 @@ from tanner.utils.mysql_db_helper import MySQLDBHelper
 
 
 def mock_config(section, value):
-    if section == 'SQLI' and value == 'host':
-        return '127.0.0.1'
-    if section == 'SQLI' and value == 'user':
-        return 'root'
-    if section == 'SQLI' and value == 'password':
-        return ''
+    config = {'host': '127.0.0.1', 'user': 'root', 'password': 'user_pass'}
+
+    return config[value]
 
 
 class TestMySQLDBHelper(unittest.TestCase):
@@ -48,11 +45,14 @@ class TestMySQLDBHelper(unittest.TestCase):
     def test_check_db_exists(self, m):
         self.expected_result = 1
 
-        async def test():
+        async def setup():
             await self.cursor.execute('CREATE DATABASE test_db')
             await self.conn.commit()
+
+        async def test():
             self.returned_result = await self.handler.check_db_exists(self.db_name)
 
+        self.loop.run_until_complete(setup())
         self.loop.run_until_complete(test())
         self.assertEqual(self.expected_result, self.returned_result)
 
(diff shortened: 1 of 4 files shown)


rjt-gupta:
The tests are passing locally, I dont understand why it fails here.

rnehra01:
Look at the logs and try to understand the errors. I'll look at it tomorrow.

rnehra01:
You need to update the Travis file to use MySql service https://docs.travis-ci.com/user/database-setup/#mysql. Also, consider the password the and username.

rjt-gupta:
> 
> 
> You need to update the Travis file to use MySql service https://docs.travis-ci.com/user/database-setup/#mysql. Also, consider the password the and username.

Just when I wanted to learn more about travis. Awesome!

rjt-gupta:
Done Finally :)

rnehra01:
What about other methods?

rjt-gupta:
Yes about that `insert_dummy_data` will be added to `test_mysqli` which I will push in some time. But the `copy_db` method creates the database in `/var/lib/mysql/*` directory. So im not sure how to test this.

rnehra01:
What is the problem?

rjt-gupta:
> 
> 
> What is the problem?

Travis build fails to enter `/var/lib/mysql` and prompts for password. Im doing more research to come up with a fix.

tanner/tests/test_mysql_db_helper.py
```diff
@@ -0,0 +1,73 @@
...
+        open('/tmp/db/test_db', 'a').close()
+
+        self.db_name = 'test_db'
+        self.expected_result = None
+        self.returned_result = None
+        self.query_map = None
+        self.handler = MySQLDBHelper()
+
+    @mock.patch('tanner.config.TannerConfig.get', side_effect=mock_config)
+    def test_check_db_exists(self, m):
```
rnehra01: I'm not sure what you're doing but you should do
1) Make a DB and test it exists
2) when no db present test method returns same result

tanner/tests/test_mysql_db_helper.py
```diff
@@ -0,0 +1,73 @@
...
+            async def delete():
+                await self.handler.delete_db(self.db_name)
+            self.loop.run_until_complete(delete())
+        else:
+            self.assertEqual(self.expected_result, self.returned_result)
+
+    @mock.patch('tanner.config.TannerConfig.get', side_effect=mock_config)
+    def test_create_query_map(self, m):
+
+        config = {
```
rnehra01: Make it more robust. Make at least 2 DBs with each at least 2 tables and different column types. Also make a table name common in the 2 DBs but different column(no of columns/column types)

tanner/utils/mysql_db_helper.py
```diff
@@ -18,7 +18,7 @@ def __init__(self):
                                       )
         return conn
 
-    async def check_db_exists(self, db_name, ):
+    async def check_db_exists(self, db_name):
         conn = await self.connect_to_db()
```
rnehra01: If it is not necessary, please don't do it.

tanner/utils/sqlite_db_helper.py
```diff
@@ -60,7 +60,7 @@ def copy_db(self, src, dst, working_dir):
         cursor.executemany("INSERT INTO " + table_name + " VALUES(" +
                            inserted_string_patt + ")", inserted_data)
 
-    def create_query_map(self, working_dir, db_name, ):
+    def create_query_map(self, working_dir, db_name):
```
rnehra01: This PR isn't related to this, please remove this change.

tanner/utils/mysql_db_helper.py
rjt-gupta: What is the use of this?

tanner/tests/test_mysql_db_helper.py
rjt-gupta: I did almost similar to this, but let me clear all this.

tanner/utils/mysql_db_helper.py
rnehra01: I'm not sure of it use either 😅

rjt-gupta: I think its a typo.

tanner/utils/sqlite_db_helper.py
rjt-gupta: I think we should remove this.

tanner/tests/test_mysql_db_helper.py
rjt-gupta: I don't think we should make complicated tests as this might confuse the actual testing function.

rnehra01: I don't think it's complicated, at least make it have more real world

rjt-gupta: Okay! I'll give it a shot!

rjt-gupta:
So, I can't directly call `copy_db` method because of `/var/lib/mysql` permissions. Any other way to test?

rnehra01:
I don't think its related to any permissions. https://github.com/rjt-gupta/tanner/blob/19b7dfb7dc938586fb7aa50c390faed1cff75867/tanner/utils/mysql_db_helper.py#L73 this is the reason of prompt since password is empty so -p flag causes a prompt, had it been a non-empty password it would have worked fine

rjt-gupta:
So, removing the -p  flag should fix the problem. Let me try.

rjt-gupta:
Worked Great!

tanner/tests/test_mysql_db_helper.py
```diff
@@ -0,0 +1,135 @@
...
+        self.handler.insert_dummy_data = AsyncMock()
+
+        async def test():
+            await self.handler.setup_db_from_config()
+
+        self.loop.run_until_complete(test())
+        assert self.handler.insert_dummy_data.called
+
+    @mock.patch('tanner.config.TannerConfig.get', side_effect=mock_config)
+    def test_copy_db(self, m):
```
rnehra01: What about checking both db are same? Maybe give it a shot https://stackoverflow.com/questions/225772/compare-two-mysql-databases

tanner/tests/test_mysql_db_helper.py
```diff
@@ -0,0 +1,135 @@
...
+                                     'TEST': [{'name': 'ID', 'type': 'INTEGER'},
+                                              {'name': 'USERNAME', 'type': 'TEXT'}]}
+
+        self.query = [
+            ['TEST_DB', "CREATE TABLE TEST (ID INTEGER PRIMARY KEY, USERNAME TEXT)", "CREATE TABLE COMMON (PARA TEXT)"],
+            ['CREDS_DB', "CREATE TABLE CREDS (ID INTEGER PRIMARY KEY, EMAIL VARCHAR(15), PASSWORD VARCHAR(15))",
+             'CREATE TABLE COMMON (NUM INTEGER )']
+        ]
+
+        async def test(data):
```
rnehra01: I think separating code use for testing from setting up code would be good. I mean don't put it in
```
await self.cursor.execute('CREATE DATABASE {db_name}'.format(db_name=data[0]))
            await self.cursor.execute('USE {db_name}'.format(db_name=data[0]))
            await self.cursor.execute(data[1])
            await self.cursor.execute(data[2])
```
in test() method. Do the same for other methods.

tanner/tests/test_mysql_db_helper.py
```diff
@@ -0,0 +1,135 @@
...
+            await self.cursor.execute('CREATE DATABASE {db_name}'.format(db_name=data[0]))
+            await self.cursor.execute('USE {db_name}'.format(db_name=data[0]))
+            await self.cursor.execute(data[1])
+            await self.cursor.execute(data[2])
+            result = await self.handler.create_query_map(data[0])
+            self.query_map.append(result)
+            await self.handler.delete_db(data[0])
```
rnehra01: Is it required?

tanner/utils/mysql_db_helper.py
```diff
@@ -68,16 +68,14 @@ def __init__(self):
             await cursor.execute('CREATE DATABASE {db_name}'.format(db_name=attacker_db))
             conn.close()
             # copy user db to attacker db
-            dump_db_cmd = 'mysqldump -h {host} -u {user} -p{password} {db_name}'
-            restore_db_cmd = 'mysql -h {host} -u {user} -p{password} {db_name}'
+            dump_db_cmd = 'mysqldump -h {host} -u {user} {db_name}'
```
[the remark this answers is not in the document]
rjt-gupta: That's the thing Initially I tried with changing mysql password in travis in my test-repo remember? It was not at all working, I guess this brings us back again to the start.

tanner/utils/mysql_db_helper.py
rjt-gupta: Alright.

tanner/utils/mysql_db_helper.py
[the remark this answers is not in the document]
rjt-gupta: And the problem is only with `copy_db` method rest are working fine.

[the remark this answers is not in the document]
rnehra01: > That's the thing Initially I tried with changing mysql password in travis in my test-repo remember? It was not at all working, I guess this brings us back again to the start.

I think the problem there was `localhost` --> `127.0.0.1`

tanner/tests/test_mysql_db_helper.py
rjt-gupta: Okay.

tanner/tests/test_mysql_db_helper.py
```diff
@@ -0,0 +1,135 @@
...
+                self.cursor = await self.conn.cursor()
+
+                # Delete DB if exists
+                self.returned_result = await self.handler.check_db_exists(self.db_name)
+                if self.returned_result == 1:
+                    await self.handler.delete_db(self.db_name)
+
+            self.loop.run_until_complete(connect())
+
+    @mock.patch('tanner.config.TannerConfig.get', side_effect=mock_config)
```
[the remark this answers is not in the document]
rjt-gupta: Yes, but whats wrong with this?

tanner/tests/test_mysql_db_helper.py
rjt-gupta: Makes sense.

tanner/tests/test_mysql_db_helper.py
rjt-gupta: I guess it is better to delete the DB after use.

tanner/tests/test_mysql_db_helper.py
```diff
@@ -67,27 +68,31 @@ def test_check_no_db_exists(self, m):
...
-        self.assertEqual(self.query_map, self.expected_result)
+        self.expected_result_creds = {'COMMON': [{'name': 'NUM', 'type': 'INTEGER'}],
+                                      'CREDS': [{'name': 'ID', 'type': 'INTEGER'}, {'name': 'EMAIL', 'type': 'TEXT'},
+                                                {'name': 'PASSWORD', 'type': 'TEXT'}]}
+
+        self.expected_result_test = {'COMMON': [{'name': 'PARA', 'type': 'TEXT'}],
+                                     'TEST': [{'name': 'ID', 'type': 'INTEGER'},
+                                              {'name': 'USERNAME', 'type': 'TEXT'}]}
+
+        self.query = [
```
rjt-gupta: Added here.

tanner/tests/test_mysql_db_helper.py
rnehra01: https://github.com/mushorg/tanner/pull/315/files#r289592042
What about this test?

rjt-gupta: As, we have added separate test for `setup_db_from_config` so I think no need to do this all again.

tanner/tests/test_mysql_db_helper.py
rjt-gupta: Done.

tanner/tests/test_mysql_db_helper.py
rnehra01: I don't understand. This methods also executes queries, so how do you whether all DBs, tables are created correctly

rjt-gupta: If the obtained query map is correct as expected then I think all tables are created correctly.

rnehra01: So in `test_query_map` the tables obtained are due to  `setup` not because of correct working of `setup_db_from_config`

rnehra01: I guess you can use query_map method here to test all the data has been created correctly.

rjt-gupta: > 
> 
> So in `test_query_map` the tables obtained are due to `setup` not because of correct working of `setup_db_from_config`

And it should be like that right? as we are only testing creation of query map.

This should be

rnehra01: Don't understand what you said here. But to be clear create_query_map tests are fine and the tests for this method are similar except here you the work of setup is done by setup_from_config method itself.

rjt-gupta: I think you might be referencing old implementation, this implementation you are talking about I have removed in latest commits.

rnehra01: Just tell me this how this test is ensuring the tables are formed correctly.

rjt-gupta: Its not. Why it is supposed to? how?

rnehra01: Because if a method makes tables then I guess its test should ensure the tables are formed correctly.

rjt-gupta:
So, `diff` is giving some mysql version specific comment but locally tests are passing. I think they are of no use here so I could grep and remove such type of output but should I do this?

rnehra01:
> So, `diff` is giving some mysql version specific comment but locally tests are passing. I think they are of not much use here and I could grep and remove such type of output but should I do this?
> 
> Refs - https://dev.mysql.com/doc/refman/8.0/en/comments.html

I think you can use querymap method to check existense of tables.

rjt-gupta:
> 
> 
> So, `diff` is giving some mysql version specific comment but locally tests are passing. I think they are of not much use here and I could grep and remove such type of output but should I do this?
> 
> Refs - https://dev.mysql.com/doc/refman/8.0/en/comments.html

No, this is im refering for `copy_db` tests as there im using `diff` to get the difference in the DB dumps. Checking the tables in `setup_db_from_config` is done.

rnehra01:
Read this https://stackoverflow.com/questions/225772/compare-two-mysql-databases , they are discussing about flags like --opt, --compact to furthur reduce comments. I don't think grepping is good.

rjt-gupta:
> 
> 
> Read this https://stackoverflow.com/questions/225772/compare-two-mysql-databases , they are discussing about flags like --opt, --compact to furthur reduce comments. I don't think grepping is good.

Okay, will try out those but I don't get why the tests are passing locally?

rnehra01:
Maybe check if travis has different version

Ravinder Nehra

rjt-gupta:
Worked Great!

tanner/tests/test_mysql_db_helper.py
rnehra01: What about this https://github.com/mushorg/tanner/pull/315/files#discussion_r287583399

rjt-gupta: Yes, tables created in `setup_db_from_config` method are checked.

rnehra01: I meant make multiple tables just like you did in createquerymap.

rjt-gupta: Okay

rjt-gupta: > 
> 
> I meant make multiple tables just like you did in createquerymap.

Done.

tanner/tests/test_mysql_db_helper.py
```diff
@@ -0,0 +1,221 @@
...
+            for table in config["tables"]:
+                await self.cursor.execute('USE test_db')
+                await self.cursor.execute('DESCRIBE {table_name}'.format(table_name=table["table_name"]))
+                result = await self.cursor.fetchall()
+                self.result.append(result)
+
+        self.loop.run_until_complete(test())
+        self.assertEqual(self.result[0], self.expected_result_test)
+        self.assertEqual(self.result[1], self.expected_result_creds)
+        assert self.handler.insert_dummy_data.called
```
rnehra01: It is called multiple times with some params. Check that.

tanner/tests/test_mysql_db_helper.py
```diff
@@ -0,0 +1,221 @@
...
+
+            for table in config["tables"]:
+                await self.cursor.execute('USE test_db')
+                await self.cursor.execute('DESCRIBE {table_name}'.format(table_name=table["table_name"]))
+                result = await self.cursor.fetchall()
+                self.result.append(result)
+
+        self.loop.run_until_complete(test())
+        self.assertEqual(self.result[0], self.expected_result_test)
+        self.assertEqual(self.result[1], self.expected_result_creds)
```
rnehra01: I think you can make expected_result a list and compare directly with result.

tanner/tests/test_mysql_db_helper.py
rjt-gupta: I tried, checking with parameters is not working that's why I used `assert called`.

tanner/tests/test_mysql_db_helper.py
rjt-gupta: Done

tanner/tests/test_mysql_db_helper.py
rnehra01: I think you have done this type of checking in a previous PR, please see how you did it there?

rjt-gupta: Done. It was something new :)

tanner/tests/test_mysql_db_helper.py
```diff
@@ -0,0 +1,219 @@
...
+            for table in config["tables"]:
+                await self.cursor.execute('USE test_db')
+                await self.cursor.execute('DESCRIBE {table_name}'.format(table_name=table["table_name"]))
+                result = await self.cursor.fetchall()
+                self.result.append(result)
+
+        self.loop.run_until_complete(test())
+        self.assertEqual(self.result, self.expected_result)
+        self.handler.insert_dummy_data.assert_called_with('CREDS', 'I,E,P', mock.ANY)
+
```
rnehra01: What about test table?

rjt-gupta: I have checked that but `setupdbfromconfig` call is only returning the last table used.

rnehra01:
Well that should not happen. Also you write code related to multiple calls
earlier. Analyze carefully because if it's not then there is
something wrong with it.

Ravinder Nehra

rjt-gupta:
Done.

tanner/tests/test_mysql_db_helper.py
rnehra01: I think you should solve these issues yourself because if you don't then it will only waste time. If I hadn't reviewed it on the same day then this issue which you already know how to fix would have taken one more day. So please try harder before raising these small issues.

PR #316, opened 2019-05-17
API tests added

rjt-gupta:


tanner/api/api.py
```diff
@@ -32,8 +32,8 @@ def __init__(self, redis_client):
                                       'cmd_exec': 0}
 
         for sess in sessions:
-            result['total_duration'] += sess['end_time'] - sess['start_time']
-            for attack in sess['attack_types']:
+            result['total_duration'] += sessions[sess]['end_time'] - sessions[sess]['start_time']
```
rnehra01: First, please submit any bug fix in a separate PR.
I don't understand this change, `sessions` is of type `list`, right?

tanner/tests/test_api.py
```diff
@@ -0,0 +1,224 @@
...
+        self.handler = Api(self.redis_client)
+
+        async def test():
+            self.returned_content = await self.handler.return_snare_info(self.uuid)
+
+        with self.assertLogs(level='ERROR') as log:
+            self.loop.run_until_complete(test())
+            self.assertIn('Can not connect to redis', log.output[0])
+
+    def test_return_session_info(self):
```
rnehra01: Add a case when `snare_uuid` is `None`. Make `return_snares` to return multiple snares id and make a `return_snare_info` method which returns different sessions for different values for snares ids.

tanner/tests/test_api.py
```diff
@@ -0,0 +1,224 @@
...
+            'user_agent': "Mozilla",
+            'peer_ip': "10.0.0.1",
+            'attack_types': "xss",
+            'possible_owners': "crawler",
+            'start_time': 148575,
+            'end_time': 148590,
+            'snare_uuid': self.snare_uuid
+        }
+
+        def mock_apply_filter(filter_name, filter_value, session):
```
rnehra01: I guess you can return false in some cases just on the basis of the session's name, like
```
if name==sess2 or sess 3:
    return False
else:
    return True
```

tanner/tests/test_api.py
```diff
@@ -0,0 +1,224 @@
...
+
+        self.expected_content = 'Invalid filter : attacktypes'
+
+        async def test():
+            self.returned_content = await self.handler.return_sessions(self.filters)
+
+        self.loop.run_until_complete(test())
+        self.assertEqual(self.expected_content, self.returned_content)
+
+    def test_apply_filter_user_agent(self):
```
rnehra01: Add some false results as well.

tanner/api/api.py
rjt-gupta: Its just a minor change. `sessions` is of type `dictionary`.

rjt-gupta:
Done 😀

tanner/tests/test_api.py
rnehra01: Also check, `apply_filter` is called with all the above filters.

tanner/tests/test_api.py
rnehra01: What about false results for this method?

tanner/api/api.py
rnehra01: I think [this](https://github.com/mushorg/tanner/blob/d55bf19b06d1430c3846896f4dfd8e1b3ce06c14/tanner/api/api.py#L53) will return a `list`. Also, I checked and it's a `list`

tanner/tests/test_api.py
rjt-gupta: On it!

tanner/api/api.py
rjt-gupta: Yup! it returns a list with the whole string `'{"sess1": {"end_time": 2.00, "start_time": 0.00 }, "sess2": {"attack_types": ["rfi"]}}'` at `0th` position. Its bcoz I made `Asyncmock` return only one element in `query_res` just to make it a bit simple.

rnehra01: I'm not sure why are you changing code to make test cases.

tanner/tests/test_api.py
rjt-gupta: I tried checking this in a loop but after running the method in `run_until_complete` it only returns the last filter of the last session not the current filters, session of the loop.

rnehra01: [This](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_has_calls) might help.

rjt-gupta:
`assert_has_calls` worked perfectly but how should I return multiple mock results based on session's name?
Also, it is getting a bit congested bcoz of asserting all this in a single test.

tanner/tests/test_api.py
```diff
@@ -0,0 +1,297 @@
...
+        filter_value = "lfi"
+
+        session = {
+            "attack_types": ["rfi", "xss"]
+        }
+
+        self.returned_content = self.handler.apply_filter(filter_name, filter_value, session)
+        self.assertFalse(self.returned_content)
+
+    def test_apply_filter_start_time(self):
```
rnehra01: I think we can add false results here too.

rnehra01:
> `assert_has_calls` worked perfectly but how should I return multiple mock results based on session's name?
> Also, it is getting a bit congested bcoz of asserting all this in a single test.

Not sure, maybe make a new test case for this `return multiple mock results based on session's name`

tanner/api/api.py
rnehra01: Any updates here?

rjt-gupta:
> 
> 
> > `assert_has_calls` worked perfectly but how should I return multiple mock results based on session's name?
> > Also, it is getting a bit congested bcoz of asserting all this in a single test.
> 
> Not sure, maybe make a new test case for this `return multiple mock results based on session's name`

I don't think we should have a separate test for that. It doesn't test any functionality anyway. Still your decision would be final. What say?

tanner/tests/test_api.py
rjt-gupta: Done.

rnehra01:
> > > `assert_has_calls` worked perfectly but how should I return multiple mock results based on session's name?
> > > Also, it is getting a bit congested bcoz of asserting all this in a single test.
> > 
> > 
> > Not sure, maybe make a new test case for this `return multiple mock results based on session's name`
> 
> I don't think we should have a separate test for that. It doesn't test any functionality anyway. Still your decision would be final. What say?

Did you test the link I sent on slack?

rjt-gupta:
> Did you test the link I sent on slack?

We don't need it fully. It was much simpler 😃

tanner/api/api.py
rnehra01: Everything looks good but I'm still not convinced here.

rjt-gupta: `sessions` is a dictionary. So, iterating over it means iterating the keys which is `string` so it raises `TypeError: string indices must be integers.`  This is just a workaround for it.

https://stackoverflow.com/questions/49695836/typeerror-string-indices-must-be-integers-python

rnehra01: No, I'm saying this is a list object.
Look here, it should return `list` object. And I have checked on my system the previous code is running fine. So I don't understand why you are calling it a dict object.

rjt-gupta: I guess its not very much clear from the code. Can you give a sample `session` object if it has to be a list?

rnehra01: Quick question first, did you change the code because you get an error while running or just by intuition that `sessions` is a `dict` object?

rjt-gupta: It was based on intution. I was having a doubt if its a list because of `json.loads`.

rnehra01: I think you should run `tannerweb` and then go to `localhost:8091/snare-stats/<snare-uuid>` which calls that function and see if it is actually properly (maybe analyze the variables types).

rjt-gupta: Roger! Ill analyze this more deeply.

rjt-gupta:
I have analyzed this and I think `sessions` can be a `list` object.
So, the structure of `sessions` might be something like - 
`sessions = [{
            "end_time": 2.00,
            "start_time": 0.00,
            "attack_types": ["lfi", "xss"]
        },
            {
                "end_time": 4.00,
                "start_time": 2.00,
                "attack_types": ["rfi", "lfi", "cmd_exec"]
            }
        ]`

@rnehra01 Can you confirm this so I can change all its occurrences?

rnehra01:
Just undo the changes related to this, the previous version was working fine for me.

rnehra01:
From the above discussion, I realized that you shouldn't mock the redis client, instead you should add data to redis accordingly and let the redis method return values based on data you added.

rjt-gupta:
> 
> 
> From the above discussion, I realized that you shouldn't mock the redis client, instead you should add data to redis accordingly and let the redis method return values based on data you added.

If we do this each and every test will process data which will take time.
But why are we testing redis client?

rjt-gupta:
> 
> 
> Just undo the changes related to this, the previous version was working fine for me.

No, I will have to mock `return_snare_info` for other tests as well so I need `sessions` object.

rnehra01:
> > Just undo the changes related to this, the previous version was working fine for me.
> 
> No, I will have to mock `return_snare_info` for other tests as well so I need `sessions` object.

Yeah, obviously do required changes but I'm talking about the source code.

rjt-gupta:
> 
> 
> > > Just undo the changes related to this, the previous version was working fine for me.
> > 
> > 
> > No, I will have to mock `return_snare_info` for other tests as well so I need `sessions` object.
> 
> Yeah, obviously do required changes but I'm talking about the source code.

Yeah! of course.

rnehra01:
> > From the above discussion, I realized that you shouldn't mock the redis client, instead you should add data to redis accordingly and let the redis method return values based on data you added.
> 
> If we do this each and every test will process data from redis which will take time.
> But why are we testing redis client?

So I don't think we should mock the library functions. For instance, in a future version if this function returns a dict object but since you have mocked it, so we won't be able to know that the new version is not compatible with our code. And our upgrading if the system breaks then we need to debug a lot to get the root cause.

rjt-gupta:
> 
> 
> > > From the above discussion, I realized that you shouldn't mock the redis client, instead you should add data to redis accordingly and let the redis method return values based on data you added.
> > 
> > 
> > If we do this each and every test will process data from redis which will take time.
> > But why are we testing redis client?
> 
> So I don't think we should mock the library functions. For instance, in a future version if this function returns a dict object but since you have mocked it, so we won't be able to know that the new version is not compatible with our code. And our upgrading if the system breaks then we need to debug a lot to get the root cause.

Got it! 
well this changes the whole `test_api.py`. So, I might need more time to sort everything out.

rjt-gupta:
Done.

rnehra01:
Also separate setup and test phase wherever possible.

tanner/tests/test_api.py
```diff
@@ -0,0 +1,330 @@
...
+        self.members = None
+        self.key = None
+
+        async def connect():
+            self.redis_client = await redis_client.RedisClient.get_redis_client()
+
+        self.loop.run_until_complete(connect())
+        self.handler = Api(self.redis_client)
+
+    def test_return_snares(self):
```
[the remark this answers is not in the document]
rjt-gupta: Done.

tanner/tests/test_api.py
```diff
@@ -0,0 +1,330 @@
...
+            'total_duration': 4.0, 'total_sessions': 2
+        }
+
+        async def test():
+            self.returned_content = await self.handler.return_snare_stats(self.snare_uuid)
+
+        self.loop.run_until_complete(test())
+        self.assertEqual(self.returned_content, self.expected_content)
+
+    def test_return_snare_info(self):
```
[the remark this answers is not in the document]
rjt-gupta: It will be better if we add the data where its required, will help to understand them better.

PR #317, opened 2019-05-22
Travis mysql setup

rjt-gupta:


tanner/config.py
```diff
@@ -21,6 +21,8 @@
                                         'php_code_injection': True, "crlf": True},
                    'SQLI': {'type': 'SQLITE', 'db_name': 'tanner_db', 'host': 'localhost', 'user': 'root',
                             'password': 'user_pass'},
+                   'MySQLI': {'type': 'MySQL', 'db_name': 'tanner_db', 'host': '127.0.0.1', 'user': 'root',
```
rnehra01: Don't change the code just to make tests. You should change the values in this `'SQLI': {'type': 'SQLITE', 'db_name': 'tanner_db', 'host': 'localhost', 'user': 'root',
                            'password': 'user_pass'}` for tests.

rjt-gupta: I thought it will be useful. No worries will keep that in mind :)

tanner/tests/test_config.py
```diff
@@ -73,8 +73,8 @@ def test_get_when_file_dont_exists(self):
             'REDIS': {'host': 'localhost', 'port': 6379, 'poolsize': 80, 'timeout': 1},
             'EMULATORS': {'root_dir': '/opt/tanner'},
             'EMULATOR_ENABLED': {'sqli': True, 'rfi': True, 'lfi': True, 'xss': True, 'cmd_exec': True},
-            'SQLI': {'type': 'SQLITE', 'db_name': 'tanner_db', 'host': 'localhost', 'user': 'root',
-                     'password': 'user_pass'},
+            'SQLI': {'type': 'SQLITE', 'db_name': 'tanner_db', 'host': '127.0.0.1', 'user': 'root',
```
rnehra01: Isn't it possible to leave it as it is and change the values of `host` and `password` in the `setUp` method?

rjt-gupta: Yeah! we can hardcode the credentials in setup directly if we can?

rjt-gupta: I don't think we can test db helpers without changing that. We can only do this for emulators tests.

rnehra01: Look [here](https://stackoverflow.com/questions/30945870/how-to-mock-patch-a-class-imported-in-another-module), you can patch the `TannerConfig` class, see if it works.

rjt-gupta:
I think let's merge the travis changes first and then we'll see the config setting in the test's PRs.

rnehra01:
Add this change in that PR itself, it will be easy to track why was is it added.

rjt-gupta:
> 
> 
> Add this change in that PR itself, it will be easy to track why was is it added.

Alright. Shall I close this PR?

PR #319, opened 2019-05-26
key format fix

rjt-gupta:
I think the key for `smembers` command should be of this format.

Refs - https://github.com/aio-libs/aioredis/blob/e8c33e39558d4cc91cf70dde490d8b330c97dc2e/tests/set_commands_test.py#L190

### DIFF
--- a/tanner/api/api.py
+++ b/tanner/api/api.py
@@ -12,7 +12,7 @@ class Api:
     async def return_snares(self):
         query_res = []
         try:
-            query_res = await self.redis_client.smembers('snare_ids', encoding='utf-8')
+            query_res = await self.redis_client.smembers('key:snare_ids', encoding='utf-8')
         except aioredis.ProtocolError as connection_error:
             self.logger.exception('Can not connect to redis %s', connection_error)
         return list(query_res)


rjt-gupta:
#316 needs this change so, please have a look at this first.

tanner/api/api.py
```diff
@@ -12,7 +12,7 @@ def __init__(self, redis_client):
     async def return_snares(self):
         query_res = []
         try:
-            query_res = await self.redis_client.smembers('snare_ids', encoding='utf-8')
+            query_res = await self.redis_client.smembers('key:snare_ids', encoding='utf-8')
```
rnehra01: The previous one is working fine for me. Did it give some error or return the wrong results?

rjt-gupta: I tried it without `key:` and error I got - `aioredis.errors.ReplyError: WRONGTYPE Operation against a key holding the wrong kind of value`. Also, while testing using `redis_client.sadd()` gave me same error without using `key:`.

rjt-gupta: Wait same thing I tried [here](https://github.com/mushorg/tanner/blob/d7b6c4b3bb0497941d12b4747fae196f5accd604/tanner/api/api.py#L45) but it works both ways. Let me check this again.

rjt-gupta: Nope `smembers` not working.

rjt-gupta:
So, I think the problem is key `snare_ids` is mapped to values of type `string` (checked through redis-cli). Everything else is of type `set`. And for type `string` we use commands such as `get`, to use `smembers`/`sadd` the key should be of type `set`.

rjt-gupta:
And Its not an issue anymore :)

PR #324, opened 2019-06-04
PHP Code Injection Emulator 

rjt-gupta:
Changed implementation using `PHPSandboxHelper`.


rnehra01:
Also add this helper in other emulators(if required) as well in this PR itself.

rjt-gupta:
I guess this would be final.

PR #325, opened 2019-06-05
PHP Object Injection Tests

rjt-gupta:


tanner/tests/test_php_object_injection.py
```diff
@@ -0,0 +1,49 @@
...
+class TestPHPCodeInjection(unittest.TestCase):
+    def setUp(self):
+        self.loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(None)
+        self.handler = PHPObjectInjection(loop=self.loop)
+        self.result = None
+        self.expected_result = None
+        self.returned_result = None
+
+    def test_scan(self):
```
rnehra01: Test with wrong regex.

rjt-gupta:
> 
> 
> What about get-results method?

From PHPSandbox Helper?

rnehra01:
https://github.com/mushorg/tanner/blob/master/tanner/emulators/php_object_injection.py#L14

Ravinder Nehra

rjt-gupta:
> 
> https://github.com/mushorg/tanner/blob/master/tanner/emulators/php_object_injection.py#L14 

It is included in `test_handle`. Should I add separate test also?

rnehra01:
Is it possible to do this without mocking phpox helper?

rjt-gupta:
> 
> 
> Is it possible to do this without mocking phpox helper?

I don't think so, there is no default phpox right?

tanner/tests/test_php_object_injection.py
rnehra01: ??

tanner/tests/test_php_object_injection.py
```diff
@@ -21,6 +21,13 @@ def test_scan(self):
         self.returned_result = self.handler.scan(payload)
         self.assertEqual(self.returned_result, self.expected_result)
 
+    def test_scan_negative(self):
```
rjt-gupta: Here.

PR #328, opened 2019-06-08
Dorks Manager Test

rjt-gupta:
All the tests are working after #330 gets merged.

But, I need help on `choose_dorks` method as while running the test `random.sample` sometimes tries to sample more items from the list of lesser size. Hence throws `ValueError: sample larger than population`.

Please have a look at the possible workaround [here](https://stackoverflow.com/questions/20861497/sample-larger-than-population-in-random-sample-python).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,6 @@ install:
   - docker-compose -f ./docker/docker-compose.yml up -d
 # command to run tests
 script:
-    - python setup.py install
     - nosetests -vv --with-coverage --cover-inclusive --cover-package=tanner
     - pycodestyle . --max-line-length=120
     - test $(docker ps -f name=tanner -q | wc --lines) -eq 5
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ install:
   - docker-compose -f ./docker/docker-compose.yml up -d
 # command to run tests
 script:
+    - python setup.py install
     - nosetests -vv --with-coverage --cover-inclusive --cover-package=tanner
     - pycodestyle . --max-line-length=120
     - test $(docker ps -f name=tanner -q | wc --lines) -eq 5
--- a/tanner/tests/test_dorks_manager.py
+++ b/tanner/tests/test_dorks_manager.py
@@ -1,6 +1,7 @@
 import unittest
 import asyncio
 import aioredis
+import random
 from unittest import mock
 from tanner import redis_client
 from tanner import config
@@ -31,7 +32,7 @@ class TestDorksManager(unittest.TestCase):
                                                self.redis_client)
 
         self.loop.run_until_complete(test())
-        assert self.redis_client.sadd.called
+        #assert self.redis_client.sadd.called
 
     def test_extract_path(self):
         self.path = 'http://example.com/index.html?page=26'
@@ -68,6 +69,16 @@ class TestDorksManager(unittest.TestCase):
 
         self.loop.run_until_complete(test())
         self.handler.push_init_dorks.assert_has_calls(calls)
+
+    def test_choose_dorks(self):
+        random.randint = mock.Mock(return_value=0)
+        self.handler.init_dorks = AsyncMock()
+
+        async def test():
+            self.returned_result = await self.handler.choose_dorks(self.redis_client)
+
+        self.loop.run_until_complete(test())
+        assert self.handler.init_dorks.called
 
     def tearDown(self):
 
--- a/tanner/tests/test_dorks_manager.py
+++ b/tanner/tests/test_dorks_manager.py
@@ -25,14 +25,14 @@ class TestDorksManager(unittest.TestCase):
         self.expected_result = None
 
     def test_push_init_dorks(self):
+        self.redis_client.sadd = AsyncMock()
 
         async def test():
             await self.handler.push_init_dorks(config.TannerConfig.get('DATA', 'dorks'), self.handler.dorks_key,
                                                self.redis_client)
-            self.returned_result = await self.redis_client.smembers(self.handler.dorks_key)
 
         self.loop.run_until_complete(test())
-        self.assertEqual(len(self.returned_result), 3562)
+        assert self.redis_client.sadd.called
 
     def test_extract_path(self):
         self.path = 'http://example.com/index.html?page=26'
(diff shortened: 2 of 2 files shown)


rnehra01:
I think you can mock `random.randint`.

rjt-gupta:
I don't understand why test are failing here, it works fine locally.

rnehra01:
I think you should check whether values are inserted correctly in redis instead of mocking library methods.

rjt-gupta:
Done.

tanner/tests/test_dorks_manager.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def test_push_init_dorks(self):
+
+        async def test():
+            await self.handler.push_init_dorks(config.TannerConfig.get('DATA', 'dorks'), self.handler.dorks_key,
+                                               self.redis_client)
+            self.returned_result = await self.redis_client.smembers(self.handler.dorks_key)
+
+        self.loop.run_until_complete(test())
+        self.assertEqual(len(self.returned_result), 3562)
```
rnehra01: Is `self.handler.dorks_key` == `3562`?
Then I guess you should substitute `self.handler.dorks_key` with some constant and check it here other wise tests will fail if there is some modification in `self.handler.dorks_key`

tanner/tests/test_dorks_manager.py
```diff
@@ -0,0 +1,90 @@
...
+        self.redis_client.sadd = AsyncMock(side_effect=aioredis.ProtocolError)
+
+        async def test():
+            await self.handler.extract_path(self.path, self.redis_client)
+
+        with self.assertLogs(level='ERROR') as log:
+            self.loop.run_until_complete(test())
+            self.assertIn('Problem with redis connection', log.output[0])
+
+    def test_init_dorks(self):
```
rnehra01: Check the case if any key doesn't exist
https://github.com/mushorg/tanner/blob/master/tanner/dorks_manager.py#L48
https://github.com/mushorg/tanner/blob/master/tanner/dorks_manager.py#L49

tanner/tests/test_dorks_manager.py
```diff
@@ -0,0 +1,90 @@
...
+        ]
+
+        async def test():
+            await self.handler.init_dorks(self.redis_client)
+
+        self.loop.run_until_complete(test())
+        self.handler.push_init_dorks.assert_has_calls(calls)
+
+    def test_choose_dorks(self):
+        random.randint = mock.Mock(return_value=0)
```
rnehra01: Please do better testing. Test return value.

.travis.yml
```diff
@@ -17,6 +17,7 @@ install:
   - docker-compose -f ./docker/docker-compose.yml up -d
 # command to run tests
 script:
+    - python setup.py install
```
rnehra01: How does this help?

tanner/tests/test_dorks_manager.py
rjt-gupta: No, `self.handler.dorks_key` is not equal to 3562. 3562 is the total length of the list which is returned by `smembers`. Here `self.handler.dorks_key` == `9445102ddf663ed0a20382c219a952ba`

tanner/tests/test_dorks_manager.py
rjt-gupta: Here its the problem if I choose values greater than zero it throws the error `ValueError: sample larger than population`. I think we don't have to test randomly chosen dorks.

.travis.yml
rjt-gupta: The problem was there is no `dorks.pickle` and `user_dorks.pickle` file which is created during `setup`.

 https://github.com/mushorg/tanner/blob/9fa7289896d194e0ca7b03faec3dfd72553271a8/tanner/dorks_manager.py#L59

tanner/tests/test_dorks_manager.py
rjt-gupta: I guess total length should remain same.

tanner/tests/test_dorks_manager.py
rjt-gupta: Done.

rnehra01: No I meant to add a case when one or both have None values.

tanner/tests/test_dorks_manager.py
rnehra01: I guess you need to dig deeper. Above you're saying the number is near 3k and here this error, so please look into it.

tanner/tests/test_dorks_manager.py
rnehra01: As far I understand this is loaded from the pickle which can change.

tanner/tests/test_dorks_manager.py
rjt-gupta: Got it.

tanner/tests/test_dorks_manager.py
rjt-gupta: Both cases are different here the problem is with `random.sample` not dorks keys.

tanner/tests/test_dorks_manager.py
```diff
@@ -25,14 +25,14 @@ def setUp(self):
...
+        self.redis_client.sadd = AsyncMock()
 
         async def test():
             await self.handler.push_init_dorks(config.TannerConfig.get('DATA', 'dorks'), self.handler.dorks_key,
                                                self.redis_client)
-            self.returned_result = await self.redis_client.smembers(self.handler.dorks_key)
 
         self.loop.run_until_complete(test())
-        self.assertEqual(len(self.returned_result), 3562)
+        assert self.redis_client.sadd.called
```
rnehra01: Don't mock library methods.

tanner/tests/test_dorks_manager.py
rnehra01: As per the errors you got, can you the actual population/size?

tanner/tests/test_dorks_manager.py
rjt-gupta: I used this to just convert it into a asyncmock object to test if `sadd` is called.

rnehra01: I understand that but I prefer the previous version of tests.

rnehra01: I understood that but I prefer the previous version.

rnehra01: Can you try passing a custom pickle with a few data samples?

rjt-gupta:
Updated both methods using custom pickle file. 😀

.travis.yml
rnehra01: Is this still required?

rjt-gupta: For now its not required, but I think its good to have this.

rnehra01:
I don't think it's required to make unnecessary changes.

rjt-gupta:
> 
> I don't think it's required to make unnecessary changes.

Okay Removed.

PR #330, opened 2019-06-08
BUG fix

rjt-gupta:
So, after banging my head for a while, the problem is with `test_config`. Its actually changing the config settings which is then eventually used by `test_dorks_manager` unlike `test_api` which already ran before it.

### DIFF
--- a/tanner/tests/test_config.py
+++ b/tanner/tests/test_config.py
@@ -2,10 +2,11 @@ import configparser
 import os
 import unittest
 
+from unittest import mock
 from tanner import config
 
 
-class TestCongif(unittest.TestCase):
+class TestConfig(unittest.TestCase):
     def setUp(self):
         config.TannerConfig.config = None
         self.d = {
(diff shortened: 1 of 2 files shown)


tanner/tests/test_config.py
```diff
@@ -41,7 +42,8 @@ def setUp(self):
 
         self.invalid_config_path = '/random/random_name'
 
-    def test_set_config_when_file_exists(self):
+    @mock.patch('tanner.tests.test_config.config')
```
rnehra01: Is it required?

rjt-gupta: Yes absolutely :)

rnehra01: But it was not needed previously.

rjt-gupta: We should patch `config` here to limit the scope to this test. Because next tests uses changed settings which should not happen. This is the reason why `test_api` works.

rjt-gupta: Do you have some other fix in mind?

rnehra01: I don't understand this code, patching is used to point one object to another. And also since you have changed the value of redis port in setUp method so I guess no more changes are required.

rjt-gupta: Patching is also used to limit the scope of the test. Other tests are also affected in `dorks_manager` which checks for certain function to be called so only changing port settings won't work. I guess patching it here does no harm to `test_config`.

PR #333, opened 2019-06-12
XXE Injection Emulator

rjt-gupta:


tanner/emulators/xxe_injection.py
```diff
@@ -0,0 +1,44 @@
...
+                   '$xml = \'%s\'; ' \
+                   '$dom = new DOMDocument(); ' \
+                   '$dom->loadXML($xml, LIBXML_NOENT | LIBXML_DTDLOAD); ' \
+                   '$creds = simplexml_import_dom($dom); ' \
+                   '$user = $creds->user; ' \
+                   '$pass = $creds->pass; ' \
+                   'echo "You have logged in as $user";' \
+                   '?> ' % code
+
+        object_injection_result = await self.helper.get_result(vul_code)
```
rnehra01: object injection?

tanner/emulators/xxe_injection.py
```diff
@@ -0,0 +1,44 @@
...
+
+        return object_injection_result
```
rnehra01: Same here.

tanner/utils/patterns.py
```diff
@@ -12,6 +12,7 @@
 PHP_CODE_INJECTION = re.compile(r'.*(;)*(echo|system|print|phpinfo)(\(.*\)).*')
 PHP_OBJECT_INJECTION = re.compile(r'(^|;|{|})O:[0-9]+:')
 CRLF_ATTACK = re.compile(r'.*(\r\n).*')
+XXE_INJECTION = re.compile(r'(^<\?xml)(\s)(version=.*)(\s)(encoding=.*)(\?>).*')
```
rnehra01: Please consider these things for regex
1. Check whether any text before keyword `<?xml` make it valid.
2. `\s` can occur multiple time, not just `\s` maybe various fields(other than `version/encoding`) can be placed here
3. I think it possible to avoid `version/encoding`.
4. Also check whether we can start directly from`<!DOCTYPE` and avoid `<?xml ... ?>`

tanner/emulators/xxe_injection.py
```diff
@@ -0,0 +1,44 @@
...
+
+class XXEInjection:
+    def __init__(self, loop=None):
+        self._loop = loop if loop is not None else asyncio.get_event_loop()
+        self.logger = logging.getLogger('tanner.xxe_injection')
+        self.helper = PHPSandboxHelper(self._loop)
+
+    async def get_injection_result(self, code):
+
+        vul_code = '<?php ' \
```
rnehra01: `'''` would be better here.

tanner/config.py
```diff
@@ -18,9 +18,11 @@
                    'REDIS': {'host': 'localhost', 'port': 6379, 'poolsize': 80, 'timeout': 1},
                    'EMULATORS': {'root_dir': '/opt/tanner'},
                    'EMULATOR_ENABLED': {'sqli': True, 'rfi': True, 'lfi': True, 'xss': True, 'cmd_exec': True,
-                                        'php_code_injection': True, 'php_object_injection': True, "crlf": True},
+                                        'php_code_injection': True, 'php_object_injection': True, "crlf": True,
+                                        "xxe_injection": True},
                    'SQLI': {'type': 'SQLITE', 'db_name': 'tanner_db', 'host': 'localhost', 'user': 'root',
                             'password': 'user_pass'},
+                   'OUT_OF_BAND': {'xxe_injection': False},
```
rnehra01: `'xxe_injection': {'OUT_OF_BAND': False}` is more suitable.

tanner/emulators/base.py
```diff
@@ -5,7 +5,8 @@
 
 from tanner import __version__ as tanner_version
 from tanner.config import TannerConfig
-from tanner.emulators import lfi, rfi, sqli, xss, cmd_exec, php_code_injection, php_object_injection, crlf
+from tanner.emulators import lfi, rfi, sqli, xss, cmd_exec, php_code_injection, php_object_injection, crlf, \
+    xxe_injection
```
rnehra01: keep it in previous line.

tanner/emulators/xxe_injection.py
rjt-gupta: Oops.

tanner/config.py
rjt-gupta: Okay

tanner/emulators/xxe_injection.py
rjt-gupta: Right.

tanner/utils/patterns.py
rjt-gupta: Yes, we can start directly from `<!DOCTYPE` and the code would work but I thought we should follow the DTD structure.

rnehra01: I think we should consider all the possibilities so relax the regex.

rjt-gupta: Okay 😅 .

rjt-gupta:
Regex - `.*<(\?xml|(\!DOCTYPE.*)).*` checks the presence of xml or doctype tags.

Working payload : `<?xml version="1.0" encoding="ISO-8859-1"?><!DOCTYPE foo [ <!ELEMENT foo ANY ><!ENTITY xxe SYSTEM "file:///etc/passwd" >]>`

tanner/utils/patterns.py
```diff
@@ -12,6 +12,7 @@
 PHP_CODE_INJECTION = re.compile(r'.*(;)*(echo|system|print|phpinfo)(\(.*\)).*')
 PHP_OBJECT_INJECTION = re.compile(r'(^|;|{|})O:[0-9]+:')
 CRLF_ATTACK = re.compile(r'.*(\r\n).*')
+XXE_INJECTION = re.compile(r'.*<(\?xml|(\!DOCTYPE.*)).*')
```
rnehra01: I think `>` can be added in the end as well.

rnehra01: Other than this it looks fine but I'll need 1-2 days to do some testing. So you can work on tests for the emulator.

rjt-gupta: Awesome :)

rjt-gupta: Should I add tests in new a PR?

rnehra01:
I guess you can add in this PR itself.

Ravinder Nehra

rjt-gupta:
Tests added.

rnehra01:
> Regex - `.*<(\?xml|(\!DOCTYPE.*)).*` checks the presence of xml or doctype tags only.
> 
> Working payload : `<?xml version="1.0" encoding="ISO-8859-1"?><!DOCTYPE foo [ <!ELEMENT foo ANY ><!ENTITY xxe SYSTEM "file:///etc/passwd" >]>`

For me adding `<scan>&xxe;</scan>` in the end produces results, can you check it again? Also, I needed to `echo $creds` in the vuln code.

tanner/emulators/xxe_injection.py
```diff
@@ -0,0 +1,46 @@
...
+    async def get_injection_result(self, code):
+
+        vul_code = '''<?php
+                        libxml_disable_entity_loader (false);
+                        $xml = \'%s\';
+                        $dom = new DOMDocument();
+                        $dom->loadXML($xml, LIBXML_NOENT | LIBXML_DTDLOAD);
+                        $creds = simplexml_import_dom($dom);
+
+                        $user = $creds->user;
```
rnehra01: Here attacker needs to guess the name of the variable `<creds><user>&xxe;</user></creds>`. Instead
```suggestion
                        $data = simplexml_import_dom($dom);
                        echo $data;
```
would be better.

rjt-gupta:
> For me adding `<creds><user>&xxe;</user></creds>` in the end produces results, can you check it again?

Yes, for phpox the whole payload I used - `<?xml version="1.0" encoding="ISO-8859-1"?><!DOCTYPE foo [ <!ELEMENT foo ANY ><!ENTITY xxe SYSTEM "file:///etc/passwd" >]><creds><user>&xxe;</user><pass>mypass</pass></creds>`. That one was for scanning, Sorry should have said before 😅 .

tanner/emulators/xxe_injection.py
rjt-gupta: To make it more intuitive we were emulating a submit form. So this can be exposed there.
Still should I change?

rnehra01: I think this should be left for honeypot user and we should focus on making it work as it is. You can add information in the docs about how and what it can emulate.

rjt-gupta: Alright.

rjt-gupta:
Changed Vuln code.

PR #334, opened 2019-06-14
Added Comments

rjt-gupta:
Areas :
- PHPSandbox Helper
- SQLite DB Helper

### DIFF
--- a/tanner/utils/sqlite_db_helper.py
+++ b/tanner/utils/sqlite_db_helper.py
@@ -12,6 +12,11 @@ class SQLITEDBHelper(BaseDBHelper):
         self.logger = logging.getLogger('tanner.sqlite_db_helper.SQLITEDBHelper')
 
     async def setup_db_from_config(self, working_dir, name=None):
+        """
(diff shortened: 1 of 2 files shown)


tanner/utils/sqlite_db_helper.py
```diff
@@ -49,6 +60,12 @@ def copy_db(self, src, dst, working_dir):
         return dst
 
     async def insert_dummy_data(self, table_name, data_tokens, cursor):
+        """
+        Inserts Dummy data in the current sqlite database
+        :param table_name (str): Table to inject dummy data
+        :param data_tokens (str): Tokens to generate dummy data, ex: 'I,L' to get ID and username data
```
rnehra01: `ex` --> `eg`. Also, I'm not sure if I,L stands for ID and username. Please confirm it.

rjt-gupta: Yes, it does - https://github.com/mushorg/tanner/blob/9fa7289896d194e0ca7b03faec3dfd72553271a8/tanner/utils/base_db_helper.py#L28

PR #341, opened 2019-07-17
Aiodocker Helper

rjt-gupta:
Changes made:
 - Cmd exec emulator
 - Lfi emulator
 - Tests


tanner/utils/aiodocker_helper.py
```diff
@@ -0,0 +1,82 @@
...
+    async def get_container(self, container_name):
+        container = None
+        try:
+            container = await self.docker_client.containers.get(container=container_name)
+
+        except aiodocker.exceptions.DockerError as server_error:
+            self.logger.exception('Error while fetching %s container %s', container_name, server_error)
+        return container
+
+    async def create_container(self, container_name, cmd=None, image=None):
```
afeena: This function, for now, is used only in tests, why you need it?

tanner/emulators/cmd_exec.py
```diff
@@ -1,20 +1,15 @@
...
+from tanner.utils import aiodocker_helper
 from tanner.utils import patterns
 
 
 class CmdExecEmulator:
     def __init__(self):
-        self.helper = docker_helper.DockerHelper()
+        self.helper = aiodocker_helper.AIODockerHelper()
 
-    async def create_attacker_env(self, session):
```
afeena: why you don't associate docker with a session anymore?

tanner/utils/aiodocker_helper.py
```diff
@@ -0,0 +1,82 @@
...
+        return container
+
+    async def execute_cmd(self, cmd, image=None):
+        execute_result = None
+        try:
+            if image is None:
+                image = self.host_image
+
+            config = {"Cmd": cmd, "Image": image}
+            container = await self.docker_client.containers.run(config=config)
```
afeena: As far as I understood, you create ll the time new containers with a random name, how you suppose to clean it?

tanner/utils/aiodocker_helper.py
```diff
@@ -0,0 +1,82 @@
...
+
+    async def delete_container(self, container_name):
+        container = await self.get_container(container_name)
+        try:
+            if container:
+                await container.delete(force=True)
+        except aiodocker.exceptions.DockerError as server_error:
+            self.logger.exception('Error while removing %s container %s', container_name, server_error)
+
+    async def close_docker(self):
```
afeena: used only in tests

tanner/emulators/lfi.py
```diff
@@ -1,28 +1,25 @@
...
         if '\x00' in file_path:
             file_path = file_path[:file_path.find('\x00')]
 
-        cmd = 'cat {file}'.format(file=shlex.quote(file_path))
-        execute_result = await self.helper.execute_cmd(container, cmd)
-        # Nulls are not printable, so replace it with another line-ender
-        execute_result = execute_result.replace('\x00', '\n')
-        return execute_result
+        cmd = ["sh", "-c", "cat {file}".format(file=shlex.quote(file_path))]
+        execute_result = await self.helper.execute_cmd(cmd)
```
afeena: since in lfi we are only read the files, it's better to create only 1 container (pass the same name) and on a shutdown of tanner remove this container

tanner/utils/aiodocker_helper.py
rjt-gupta: Yes, because with aiodocker we can execute commands in both ways, by using `client.containers.create` and with `docker_client.containers.run` as well. But, its good to have if we need to create containers at some point then it can help :)

tanner/utils/aiodocker_helper.py
rjt-gupta: Should I close the client directly in tests and remove this?

tanner/emulators/cmd_exec.py
rjt-gupta: For the above reason, I only followed one method to execute commands (`containers.run`) :) so haven't used a container like the previous implementation.

tanner/utils/aiodocker_helper.py
rjt-gupta: For now these containers have to be deleted manually because of the risk other containers can also be deleted if we try to delete them all. Do you have something in mind?

tanner/emulators/lfi.py
rjt-gupta: As such we don't really need to create a container before to execute some command.

afeena: Run method creates the containers, yes, but since you don't provide any name, it will create each time a new container, which is not ok.

tanner/utils/aiodocker_helper.py
afeena: Previously we had a procedure to delete containers in session routine, the idea is to have an associated container with session and delete it by name while deleting the session

rjt-gupta: Ahh! Yes, then in this case we should use create method to run commands in emulators right?

tanner/emulators/cmd_exec.py
rjt-gupta: You're right, got your point.

tanner/utils/aiodocker_helper.py
afeena: You can use `create`, `get` and `start` methods
or
You can still create the containers with `run` and delete it right after the execution.

afeena: I don't know what is better in terms of architecture and performance, you can test it :)

rjt-gupta: I think the second method works best.

afeena: The only problem I can imagine with a second approach is a (possible) inconsistency in the sequence of requests. And also, have you measured what is faster: always create and delete new one or get existing and start it? I would be happy to see such compariosn

tanner/utils/aiodocker_helper.py
afeena: I think in test yes and on tanner shutdown either

tanner/utils/aiodocker_helper.py
rjt-gupta: So, I analyzed the performance of both methods - (averaged multiple runs)

1. **create_or_replace:** gets the container and replaces new config or creates a new one + starting the container - **1.4206 sec**  (deleting time and logs output time has to be added)

2. **run + delete:** creates a new container and deleting [ had to force :( ] container + checking output using logs- **1.5324 sec**

tanner/utils/aiodocker_helper.py
rjt-gupta: Done.

tanner/utils/aiodocker_helper.py
rjt-gupta: I think it's not a significant performance difference, but the method you will prefer would be final :)

PR #344, opened 2019-07-26
Template Injection emulator

rjt-gupta:


tanner/emulators/template_injection.py
```diff
@@ -0,0 +1,68 @@
...
+                execute_result = execute_result[2:-2]
+
+        elif patterns.TEMPLATE_INJECTION_MAKO.match(payload):
+
+            mako_template = 'from mako.template import Template\n' \
+                            'mako_template = Template("""%s""")\n' \
+                            'template_injection_result = mako_template.render()\n' \
+                            'print(template_injection_result)' % payload
+
+            cmd = ["python3", "-c", mako_template]
```
afeena: you duplicate the code, the `cmd` and `execute result` logic is similar for both templates

tanner/emulators/template_injection.py
```diff
@@ -0,0 +1,68 @@
...
+class TemplateInjection:
+
+    def __init__(self, loop=None):
+        self._loop = loop if loop is not None else asyncio.get_event_loop()
+        self.logger = logging.getLogger('tanner.template_injection')
+        self.docker_helper = AIODockerHelper()
+
+    async def get_injection_result(self, payload):
+        execute_result = None
+        github_remote_path = 'https://raw.githubusercontent.com/mushorg/tanner/master/docker/tanner/' \
```
afeena: better to do it via config value (and probably it's the point when you can provide alternative source)

afeena:
Please, close another pr if you want only this to be merged. And resolve test problems

tanner/emulators/template_injection.py
```diff
@@ -0,0 +1,68 @@
...
+                             'template_injection/Dockerfile'
+
+        # Build the custom image
+        await self.docker_helper.setup_host_image(
+            remote_path=github_remote_path, tag='template_injection:latest')
+
+        if patterns.TEMPLATE_INJECTION_TORNADO.match(payload):
+
+            tornado_template = 'import tornado\n' \
```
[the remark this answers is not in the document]
rjt-gupta: Sure,

tanner/emulators/template_injection.py
rjt-gupta: Yes this I did this intentionally bcoz in tornado engine the execute result is having an extra string `"b''"` that I have to remove but in mako the output is ok.

But, I will try to remove redundancy more :)

tanner/emulators/template_injection.py
rjt-gupta: Perfect.

rjt-gupta:
> 
> 
> Please, close another pr if you want only this to be merged. And resolve test problems

Yes, that is I was trying to resolve and apparently the `run` method in aiodocker tries to pull the image `template_injection:latest` which doesn't exist in our case.

rjt-gupta:
> What about jinja2? What is the problem with that engine?

Only basic payloads were working like `{{7*7}}`, I tried but haven't found a way to execute more complex payloads securely.

Another engine I tried was `twig`, here basic payloads worked like charm but there was no output for complex payloads. I can't the problem with this bcoz there were no errors from php sandbox whatsoever it just returns empty output.

tanner/emulators/template_injection.py
rnehra01: I think you can make changes in tornado/mako to return results in a standard way.

rjt-gupta: Yes, im on it just need a little more time :)

tanner/emulators/template_injection.py
```diff
@@ -0,0 +1,65 @@
...
+        # Build the custom image
+        await self.docker_helper.setup_host_image(
+            remote_path=github_remote_path, tag='template_injection:latest')
+
+        if patterns.TEMPLATE_INJECTION_TORNADO.match(payload):
+
+            with open('tanner/files/engines/tornado.py', 'r') as f:
+                tornado_template = f.read() % payload
+
+            cmd = ["python3", "-c", tornado_template]
```
rnehra01: Maybe you can just write `python` and install only python3 in dockerfile so that while upgrading multiple changes are not required.

tanner/emulators/template_injection.py
```diff
@@ -0,0 +1,66 @@
...
+    async def get_injection_result(self, payload):
+        execute_result = None
+
+        try:
+
+            # Build the custom image
+            await self.docker_helper.setup_host_image(
+                remote_path=self.remote_path, tag='template_injection:latest')
+
+            if patterns.TEMPLATE_INJECTION_TORNADO.match(payload):
```
rnehra01: Is it possible to remove the b'' inside the engine itself.

rjt-gupta: I don't think so bcoz we are just rendering the template in the engine and I guess its tornado doing this itself.

rjt-gupta:
Done finally :)

tanner/emulators/template_injection.py
rjt-gupta: Good idea! made the changes in the emulator and dockerfile already installs `python3` :)

tanner/emulators/template_injection.py
```diff
@@ -0,0 +1,62 @@
...
+        execute_result = None
+        work_dir = os.getcwd()
+
+        # Build the custom image
+        await self.docker_helper.setup_host_image(
+            remote_path=self.remote_path, tag='template_injection:latest')
+
+        if patterns.TEMPLATE_INJECTION_TORNADO.match(payload):
+
+            with open(work_dir + '/tanner/files/engines/tornado.py', 'r') as f:
```
afeena: tanner can be started at any directory, the file might be not available, you can copy it into the data folder during installation and use the path from config mayb

rjt-gupta: Perfect, made the changes but can you please have a look at this Travis error log?

rnehra01:
Are tests passing locally?

rjt-gupta:
> 
> 
> Are tests passing locally?

Yes.

afeena:
I guess the problem might be that building an image and running docker takes time, could you just mock it?

afeena:
I have exactly the same problem locally, on `test_handle_mako (tanner.tests.test_template_injection.TestTemplateInjection) ... ` if just frezees

afeena:
In the execution this code causes the problem:
``` with open(work_dir, 'r') as f:
                mako_template = f.read().format(payload)```

`work_dir` is a magick mock object, so reading fails and tests are frozen

rjt-gupta:
Done :)

I dont know why I am not able to reproduce this locally, everything worked perfectly.

afeena:
Have you tested it? When I try to test it I got HTTP 504 error. And I do not see any logs. Could you please verify

path
`/index.php?p={{7*7}}`

rjt-gupta:
> 
> 
> Have you tested it? When I try to test it I got HTTP 504 error. And I do not see any logs. Could you please verify
> 
> path
> `/index.php?p={{7*7}}`

Okay So, this problem is bcoz of using python instead of python3, but the dockerfile itself is installing only python 3.

Can you try it now?

afeena:
> Can you try it now?

I still have the same behavior. Can you reproduce it?

rjt-gupta:
> I still have the same behavior. Can you reproduce it?

Is seems to work fine for me:

tornado - 

tornaado

tornado


mako - 

mako

afeena:
> Can you try `setup.py install` for new build?

:) For sure

afeena:
Docker container is created, but not deleted, so somewhere in between some problem exists

afeena:
Further investigation shows that the problem in the function
` await self.docker_client.images.build(**params)`

Might be some docker related issue
https://github.com/aio-libs/aiodocker/blob/master/tests/test_images.py#L11 look at the API requirements

rjt-gupta:
> 
> 
> Further investigation shows that the problem in the function
> ` await self.docker_client.images.build(**params)`
> 
> Might be some docker related issue
> https://github.com/aio-libs/aiodocker/blob/master/tests/test_images.py#L11 look at the API requirements

I guess for now `aiodocker` only supports `18.03.1` and `17.12.1` versions only. But we are using latest `18.09`.

https://github.com/aio-libs/aiodocker/blob/88d0285ddba8e606ff684278e0a831347209189c/tests/conftest.py#L13

rjt-gupta:
> **upd:** Changing the API from 1.26 to 1.40 didn't help

Should I try tar method then?

rjt-gupta:
Done. 

Added tests, `order=4` logic.

PR #345, opened 2019-08-02
Improved Logging in emulators

### LINKED ISSUE #261: No correlation between log and downloaded file
The other day I have been noticing that there does not seem to be a correlation between downloaded files (i.e. wrt to a RFI) and the corresponding log entry. There should be an addition to the log regarding the written file alongside with a hash.

rjt-gupta:
Fixes #261

### DIFF
--- a/tanner/emulators/rfi.py
+++ b/tanner/emulators/rfi.py
@@ -75,7 +75,7 @@ class RfiEmulator:
     async def get_rfi_result(self, path):
         rfi_result = None
         await asyncio.sleep(1, loop=self._loop)
-        self.logger.info('Downloading the file has started..')
+        self.logger.info('Downloading the file has started from %s', path)
         file_name = await self.download_file(path)
         if file_name is None:
             return rfi_result
--- a/tanner/emulators/rfi.py
+++ b/tanner/emulators/rfi.py
@@ -10,7 +10,6 @@ from concurrent.futures import ThreadPoolExecutor
 import aiohttp
 import yarl
 
-from tanner import config
 from tanner.utils.php_sandbox_helper import PHPSandboxHelper
 from tanner.utils import patterns
 
@@ -50,6 +49,7 @@ class RfiEmulator:
                 tmp_filename = url.name + str(time.time())
                 file_name = hashlib.md5(tmp_filename.encode('utf-8')).hexdigest()
                 with open(os.path.join(self.script_dir, file_name), 'bw') as rfile:
+                    self.logger.debug('Saving the RFI script %s', os.path.join(self.script_dir, file_name))
                     rfile.write(data.encode('utf-8'))
         return file_name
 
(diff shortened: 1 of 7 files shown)


tanner/emulators/rfi.py
```diff
@@ -74,6 +75,7 @@ def download_file_ftp(self, url):
     async def get_rfi_result(self, path):
         rfi_result = None
         await asyncio.sleep(1, loop=self._loop)
+        self.logger.info('Downloading the file has started..')
```
afeena: add the source like `from <path>`

tanner/emulators/xxe_injection.py
```diff
@@ -55,8 +56,10 @@ def scan(self, value):
 
         result = await self.get_injection_result(attack_params[0]['value'])
         if not result or 'stdout' not in result:
+            self.logger.exception('Error while getting the injection results from php sandbox..')
```
afeena: show the output of sandbox?

rjt-gupta: Output of sandbox is not neat, a lot of random stuff can be there. I dont think it will be good?

tanner/emulators/rfi.py
rjt-gupta: cool :)

tanner/emulators/sqli.py
```diff
@@ -46,8 +48,10 @@ def map_query(self, attack_value):
                 error_result = 'You have an error in your SQL syntax; check the manual\
                                 that corresponds to your MySQL server version for the\
                                 right syntax to use near {} at line 1'.format(attack_value['id'])
+                self.logger.exception('Error while executing MySQL %s', error_result)
```
afeena: why it's an exception? it's emulated error, not the tanner one :) maybe make it debug? also `error_result` can be logged without code duplication

tanner/emulators/sqli.py
```diff
@@ -46,8 +48,10 @@ def map_query(self, attack_value):
...
             else:
                 error_result = 'SQL ERROR: near {}: syntax error'.format(attack_value['id'])
+                self.logger.exception('Error while executing SQL %s', error_result)
```
afeena: same here, we should log exception only in case it's tanner exception and not something we want emulate

tanner/emulators/xxe_injection.py
afeena: Yeah, right

afeena:
Are you considering add more logging in other emulators? e.g. like this cases https://github.com/mushorg/tanner/blob/master/tanner/emulators/mysqli.py#L34

tanner/emulators/sqli.py
rjt-gupta: Got it!

tanner/emulators/sqli.py
rjt-gupta: Okay

rjt-gupta:
> 
> Also I looked briefly into the emulators, sometimes there is no any checks for parameters, maybe you can also include with together with logging

Can you please point me one spot?

afeena:
@rjt-gupta There is a bunch of assumptions like this about data 
https://github.com/mushorg/tanner/blob/master/tanner/emulators/crlf.py#L13
so here you can see might be a KeyError e.g. 

We can think about it and if we want to change it - make a new pr :)

PR #355, opened 2020-02-24
Fixes #354 -- Updated the README.md

### LINKED ISSUE #354: Misleading Installation guide in README.md
The [getting started](https://github.com/mushorg/tanner#getting-started) section in the README is a bit confusing for newcomers, I think.

It should be clearly mentioned that `redis` and `php sandbox` both are compulsory for tanner to function while docker may be an optional alternative if someone is unable to set up the environment on their machine.

Also, for me, Python3.5.2 was causing some problems with the dependencies especially `aiohttp` whose lastest version requires python3.6.x and above.

Parth1811:
Changed the README.md with a more user-friendly installation guide,and updated the python version as the older versioni is not supported anymore.


README.md
```diff
@@ -26,64 +29,64 @@ Basic Concept
...
 
-Run ``sudo tannerapi``
+### Install and run TANNER using docker container
 
-### Run Tanner WebUI
+  In case you want to run the TANNER service using docker or facing any problem
+  in setting up TANNER on your machine, you can follow these steps.
 
-Run ``sudo tannerweb``
+#### Setup Docker
```
afeena: This step should be before tanner setup, since tanner using docker inside. It will not work without docker

Parth1811: I have made the required changes

README.md
```diff
@@ -26,64 +29,70 @@ Basic Concept
...
-### Setup PHP Sandbox
+2. Run ``redis-server`` (to start it on `localhost` with default `port`)
 
+#### Step 2: Setup PHP Sandbox
 
-1. For PHP Sandbox setup, see sandbox [manual](https://github.com/mushorg/phpox)
-2. In PHP Sandbox directory, run sandbox: ``sudo python3 sandbox.py``
+1. Clone the repo using:  ``git clone https://github.com/mushorg/phpox.git`` (preferably in your home directory).
+2. Go to PHP Sandbox directory: ``cd phpox``
+2. Run sandbox: ``sudo python3 sandbox.py``
```
afeena: phpox should be setup before running ``sudo python3 sandbox.py``, so the first step in old version is really the first

README.md
```diff
@@ -26,64 +29,70 @@ Basic Concept
...
 
-### Setup Docker
+For more help on PHP Sandbox setup, see sandbox [manual](https://github.com/mushorg/phpox)
 
+#### Step 3: Setup Docker
 
-1. Install [docker](https://docs.docker.com/engine/installation/linux/ubuntu/)
-2. Pull the required image to use [default : ``busybox:latest``]
```
afeena: Why you've deleted this from REDME?

README.md
Parth1811: Sorry, my bad. There is only `make` command to setup phlox according to the README; I was going to add that.

Apart from that, there isn't anything needed to setup phpox, I think. So shifted the manual link to the bottom.

README.md
Parth1811: I have replaced the 1st step with the direct command, which will install the docker (which is given in the link), and shifted the link below if needed for troubleshooting.

Regarding the second point, I don't know what exactly it meant, and I was able to setup tanner on docker just using those commands which I have listed, hence deleted it.

afeena: I was talking about the second point :) I was trying to figure out why we had it explicitly, while host image is defined in config and managed by our docker_helper :)

Parth1811: Yeah exactly its redundant

README.md
```diff
@@ -26,64 +29,71 @@ Basic Concept
...
-### Setup PHP Sandbox
+2. Run ``redis-server`` (to start it on `localhost` with default `port`)
 
+#### Step 2: Setup PHP Sandbox
 
-1. For PHP Sandbox setup, see sandbox [manual](https://github.com/mushorg/phpox)
-2. In PHP Sandbox directory, run sandbox: ``sudo python3 sandbox.py``
+1. Clone the repo using:  ``git clone https://github.com/mushorg/phpox.git`` (preferably in your home directory).
+2. Go to PHP Sandbox directory: ``cd phpox``
+3. Build sandbox: ``make``
```
afeena: before make, you have to setup BFR first, I prefer to explicitly mention that phpox need setup and refer to the original phpox documentation

README.md
```diff
@@ -26,64 +29,70 @@ Basic Concept
...
-
-### Setup PHP Sandbox
+2. Run ``redis-server`` (to start it on `localhost` with default `port`)
 
+#### Step 2: Setup PHP Sandbox
 
 1. For PHP Sandbox setup, see sandbox [manual](https://github.com/mushorg/phpox)
-2. In PHP Sandbox directory, run sandbox: ``sudo python3 sandbox.py``
+2. Go to PHP Sandbox directory: ``cd phpox``
+3. Run sandbox: ``sudo python3 sandbox.py``
```
afeena: My suggestion is to put 3. Setup sandbox: see <link>
Because now if I follow your guide, I will go to the directory, try to run `sudo python3 sandbox.py` and it will fail with `no such file or directory`. It's confusing.

I am sorry for so many fixes, but since it's readme it's super important how it looks like :)

Parth1811: In that case, I would suggest we should keep the old set of instructions for the setup of `phpox' ?

It's okay, I understand its necessary to get the README correct :)

afeena: Yeah you can keep the old one

PR #358, opened 2020-02-26
Fixed issue with installing dependencies

puneet29:
The dependencies need to be installed in superuser mode else it results in errors.


mzfr:
In my opinion, rather than changing the `installation command` to run with `sudo` we should add a command to build the `virtual env` and then run the installation command inside the venv without any `sudo`

Parth1811:
@puneet29  I have already added a PR with a better version of README in #355, please have a look I have added your suggestion as well.
Also, I agree with @mzfr that there should be a virtual environment for TANNER, I would try to push a PR for that soon.

puneet29:
Yeah, I suggested for the virtual environment too @mzfr @Parth1811 
Also, I will close this pr once yours is merged.

afeena:
@puneet29 I've merged #355, please update your changes accordingly, thanks!

puneet29:
@afeena I've updated the pull request as requested.

PR #363, opened 2020-03-03
Update docker to accomodate for latest changes in master

t3chn0m4g3:
@afeena Please test :)

### DIFF
--- a/docker/phpox/Dockerfile
+++ b/docker/phpox/Dockerfile
@@ -1,10 +1,11 @@
 FROM alpine
-
+#
 # Include dist
 ADD dist/ /root/dist/
-
+#
 # Install packages
-RUN apk -U --no-cache add \
+RUN sed -i 's/dl-cdn/dl-2/g' /etc/apk/repositories && \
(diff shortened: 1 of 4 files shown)


docker/phpox/Dockerfile
```diff
@@ -1,10 +1,11 @@
...
-
+FROM alpine:3.10
+#
 # Include dist
 ADD dist/ /root/dist/
-
+#
 # Install packages
-RUN apk -U --no-cache add \
+RUN sed -i 's/dl-cdn/dl-2/g' /etc/apk/repositories && \
```
mainnika: just wondering, why you prefer dl-2 ?

afeena:
@t3chn0m4g3 Works fine, thanks!

docker/phpox/Dockerfile
t3chn0m4g3: At around 14:00 UTC building is so slow from `dl-cdn` that only switching to `dl-2` works reasonably fast :bowtie:

PR #371, opened 2020-03-09
Use YAML format for tanner configuration

### LINKED ISSUE #366: Use YAML to define configs
I suggest using yaml for configs to make it more convenient

mzfr:
Fixes #366 

Few things might fail but I am working on this. Made the PR so you(@afeena) can take a look and give your input about the way it's done.


mzfr:
@afeena Finally all the tests are passing so please review it.

config.ini
```diff
@@ -0,0 +1,2 @@
+[API]
```
afeena: why you need this file? Don't we want to move to yml?

tanner/config.py
```diff
@@ -1,92 +1,66 @@
...
-
-class TannerConfig():
+
+def read_config(path):
+    with open(path, 'r') as f:
+        config_values = yaml.load(f, Loader=yaml.FullLoader)
+    return config_values
+
+
+class Meta(type):
```
afeena: why you need meta class?

mzfr: I was using Meta class to set the default value of the config within the class.
But now I realized it would be just fine if we can stick to the old model of config i.e using a GLOBAL variable to hold the default values.

config.ini
mzfr: Sorry, this was by mistake. Since we are going to use YAML config file I've removed this from the source.

tanner/config.py
```diff
@@ -1,92 +1,52 @@
...
-                   'HPFEEDS': {'enabled': False, 'HOST': 'localhost', 'PORT': 10000, 'IDENT': '', 'SECRET': '',
-                               'CHANNEL': 'tanner.events'},
-                   'LOCALLOG': {'enabled': False, 'PATH': '/tmp/tanner_report.json'},
-                   'CLEANLOG': {'enabled': False},
-                   'REMOTE_DOCKERFILE': {'GITHUB': "https://raw.githubusercontent.com/mushorg/tanner/master/docker/"
-                                                   "tanner/template_injection/Dockerfile"},
-                   'SESSIONS': {"delete_timeout": 300}
-                   }
+
+def read_config(path):
```
afeena: you can make this function also a class function

tanner/config.py
```diff
@@ -1,92 +1,52 @@
...
+    config_values = {}
+    with open(path, 'r') as f:
+        config_values = yaml.load(f, Loader=yaml.FullLoader)
+    return config_values
+
+
+DEFAULT_CONFIG = read_config("/opt/tanner/data/config.yaml")
```
afeena: why do not read default while setting config?

tanner/config.py
```diff
@@ -1,92 +1,52 @@
...
-                convert_type = type(config_template[section][value])
-                if convert_type is bool:
-                    res = TannerConfig.config.getboolean(section, value)
-                else:
-                    res = convert_type(TannerConfig.config.get(section, value))
-            except (configparser.NoOptionError, configparser.NoSectionError):
-                LOGGER.warning("Error in config, default value will be used. Section: %s Value: %s", section, value)
-                res = config_template[section][value]
-
+                res = TannerConfig.config[section][value]
```
afeena: instead of using if/try you can use noly try and catch also TypeError (`case TannerConfig.config is None`)

or

you can merge together default config and user config

afeena:
Looks good :) I tested it a little bit and it works well, just some minor comments :)

mzfr:
@afeena The suggested changes have been made.

afeena:
If I try to set only one particular value through config, I get an error:

```
  File "/usr/local/bin/tanner", line 4, in <module>
    __import__('pkg_resources').run_script('Tanner==0.6.0', 'tanner')
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 661, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 1448, in run_script
    exec(script_code, namespace, namespace)
  File "/usr/local/lib/python3.7/site-packages/Tanner-0.6.0-py3.7.egg/EGG-INFO/scripts/tanner", line 35, in <module>
  File "/usr/local/lib/python3.7/site-packages/Tanner-0.6.0-py3.7.egg/EGG-INFO/scripts/tanner", line 30, in main
  File "/usr/local/lib/python3.7/site-packages/Tanner-0.6.0-py3.7.egg/tanner/server.py", line 31, in __init__
  File "/usr/local/lib/python3.7/site-packages/Tanner-0.6.0-py3.7.egg/tanner/emulators/base.py", line 19, in __init__
KeyError: 'rfi'
```

my config file:
```
EMULATOR_ENABLED: 
  lfi: False 
```

tanner/config.py
```diff
@@ -1,92 +1,52 @@
...
-                   'HPFEEDS': {'enabled': False, 'HOST': 'localhost', 'PORT': 10000, 'IDENT': '', 'SECRET': '',
-                               'CHANNEL': 'tanner.events'},
-                   'LOCALLOG': {'enabled': False, 'PATH': '/tmp/tanner_report.json'},
-                   'CLEANLOG': {'enabled': False},
-                   'REMOTE_DOCKERFILE': {'GITHUB': "https://raw.githubusercontent.com/mushorg/tanner/master/docker/"
-                                                   "tanner/template_injection/Dockerfile"},
-                   'SESSIONS': {"delete_timeout": 300}
-                   }
+
+class ReadConfig():
```
afeena: maybe it's better to move read_config in tanner config class? Have a separate class for only one function which is logically connected with other general class seems redundant for me

Parth1811:
Hey, won't this make the old config files obsolete?

mzfr:
@afeena Yes, that is because in `base.py` we are using `TannerConfig.get_section` to grab `EMULATOR_ENABLED` section, which in our case is present but doesn't have all the values.

One solution is that we have a kind of check function, like if the section doesn't have all the values then the missing values will be added from the default config. 

@Parth1811 Yes, I think that was the idea.

afeena:
@mzfr my point is, that it would be nice to have this property, just e.g. I want to switch off one particular emulator, so I don't want to define al others as true. I suggest you merge two dict, in this case, you can keep only one copy of config and don't do additional checks.

And as @Parth1811 mention, the old config will be not supported anymore, so it would be nice to catch the exception while trying to load yml config (in case someone wants to load ini file)

mzfr:
@afeena I've made the changes that we discussed. Also, I have updated the config docs.

Let me know if something is still missing out.

tanner/config.py
```diff
@@ -1,92 +1,44 @@
...
 
 
 class TannerConfig():
     config = None
 
+    @staticmethod
+    def read_config(path):
+        config_values = {}
+        with open(path, 'r') as f:
+            config_values = yaml.load(f, Loader=yaml.FullLoader)
```
afeena: ,I think it would be nice anyway to catch the exception instead of checking the ending of filename here https://github.com/mushorg/tanner/pull/371/files#diff-f428ee775d6702ef490b2120c5ef2ae5R23, because e.g. I've never used .ini for configs, I used `.config` :)

docs/source/config.rst
```diff
@@ -73,23 +75,100 @@ There are 8 different sections :
...
-     'EMULATORS': {'root_dir': '/opt/tanner'},
-     'EMULATOR_ENABLED': {'sqli': 'True', 'rfi': 'True', 'lfi': 'True', 'xss': 'True', 'cmd_exec': 'True'},
-     'SQLI': {'type':'SQLITE', 'db_name': 'tanner_db', 'host':'localhost', 'user':'root', 'password':'user_pass'},
-     'DOCKER': {'host_image': 'busybox:latest'},
-     'LOGGER': {'log_file': '/opt/tanner/tanner.log'},
-     'MONGO': {'enabled': 'False', 'URI': 'mongodb://localhost'},
-     'LOCALLOG': {'enabled': 'False', 'PATH': '/tmp/tanner_report.json'},
-     'CLEANLOG': {'enabled': 'False'}
-     }
+  DATA:
```
afeena: You can use comments to describe a little bit sections :)
I find this example quite good :)
https://opennmt.net/OpenNMT-tf/configuration.html

afeena:
@mzfr Awesome! It's ready to merge, since it contains breaking changes, what do you think if we merge it into develop first together with upcoming docker and dir changes and then prepare new release?

mzfr:
@afeena yeah that would be a good thing to do.

afeena:
@mzfr could you please then change destination to develop branch?

mzfr:
@afeena Done. And luckily I don't have to rebase :)

afeena:
@mzfr great, thanks!

PR #373, opened 2020-03-10
asyncio.create_task() does not work in python version less than 3.7 fixed #372

### LINKED ISSUE #372: asyncio.create_task() does not work in python version less than 3.7
In server.py file, inside the function `async def start background delete`, we have used asyncio.create_task(), which is not supported in python version less than 3.7.

steve7158:
For its solution i have checked the versionof python and accordingly used `asyncio.create_task()`(for python version 3.7 or more) and `asyncio.ensure_future()`(for python version less than 3.7)


afeena:
@steve7158 https://docs.python.org/3.7/library/asyncio-task.html#asyncio.create_task

If you can use only one function for everything - just go for it

steve7158:
Then i guess i should just change create_task to ensure_future

Issue #359, opened 2020-02-27
jinja2 exception in Tanner Web


mzfr:
When I visit `/snares` I can see two snare UUID is one I can see snare stats but in another when I click on the `snare-stats` option it gives `500 Internal server error` and I see the following traceback on the terminal:

```
➜ sudo tannerweb
======== Running on http://0.0.0.0:8091 ========
(Press CTRL+C to quit)
Error handling request
Traceback (most recent call last):
  File "/home/mzfr/dev/tanner/venv/lib/python3.8/site-packages/aiohttp/web_protocol.py", line 418, in start
    resp = await task
  File "/home/mzfr/dev/tanner/venv/lib/python3.8/site-packages/aiohttp/web_app.py", line 458, in _handle
    resp = await handler(request)
  File "/home/mzfr/dev/tanner/venv/lib/python3.8/site-packages/aiohttp_jinja2/__init__.py", line 101, in wrapped
    response = render_template(template_name, request, context,
  File "/home/mzfr/dev/tanner/venv/lib/python3.8/site-packages/aiohttp_jinja2/__init__.py", line 73, in render_template
    text = render_string(template_name, request, context, app_key=app_key)
  File "/home/mzfr/dev/tanner/venv/lib/python3.8/site-packages/aiohttp_jinja2/__init__.py", line 64, in render_string
    text = template.render(context)
  File "/home/mzfr/dev/tanner/venv/lib/python3.8/site-packages/jinja2/environment.py", line 1090, in render
    self.environment.handle_exception()
  File "/home/mzfr/dev/tanner/venv/lib/python3.8/site-packages/jinja2/environment.py", line 832, in handle_exception
    reraise(*rewrite_traceback_stack(source=source))
  File "/home/mzfr/dev/tanner/venv/lib/python3.8/site-packages/jinja2/_compat.py", line 28, in reraise
    raise value.with_traceback(tb)
  File "tanner/web/templates/snare-stats.html", line 1, in top-level template code
    {% extends "base.html" %}
  File "tanner/web/templates/base.html", line 9, in top-level template code
    {% block content%}
  File "tanner/web/templates/snare-stats.html", line 21, in block "content"
    {% for key, val in snare_stats.attack_frequency.items() %}
  File "/home/mzfr/dev/tanner/venv/lib/python3.8/site-packages/jinja2/environment.py", line 471, in getattr
    return getattr(obj, attribute)
[...]

PR #376, opened 2020-03-12
Check before displaying the value.

mzfr:
If in some case the dict snare_stats is missing value then
that would result in jinja exception

Fixes: #359


afeena:
@mzfr please resolve problems with tests

mzfr:
@afeena I've updated the tests since there was some issue with `\n\t` things.

PR #377, opened 2020-03-12
Fixes #360 -- Install tanner using in a virtualenv

### LINKED ISSUE #360: Make a virtual environment for tanner
Please move to a virtual environment for running tanner. This will require changing the installation directory of tanner from /opt/tanner. @glaslos @afeena can suggest what would be the right way to do the same.

Parth1811:
Step to install:-

1. Make a virtual environment
2. `pip install -r requrement.txt`
3. Go to the Tanner DIrectory
4. `pip install -e .`
5. Run `tanner`

Note:- this is not a full implementation there are some changes and tests to be worked on, just wanted to get a review on this.


tanner/config.py
```diff
@@ -3,22 +3,23 @@
 import os
 import sys
 
+TANNER_DIR = os.path.expanduser("~/.tanner")
```
afeena: And what if I want to setup TANNER_DIR? I should always change it in the code?

Parth1811: we can have an option while setup to change the default directory.........should we do this? is there a need for this?

afeena: if it's possible and easy to do, I definitely prefer to have a control on folders usage :)

afeena:
@Parth1811 Could you please update me with the status of this PR? :)

afeena:
Closed since no updates

PR #379, opened 2020-03-19
Update README.md

glaslos:


README.md
```diff
@@ -23,13 +23,13 @@ Basic Concept
...
 - Serve dorks.
-- Adopt and change the responses.
+- Emulate vulnerabilities and provide responses.
 
 
 Getting Started
 ---------------
 
-- You need Python3.6 and above for installing tanner.
+- You need Python3.7 and above for installing tanner.
```
afeena: we don't want to support python 3.6 anymore?

glaslos: 3.6 is no longer supported: https://en.wikipedia.org/wiki/History_of_Python#Table_of_versions but your call if you want to keep it at 3.6

afeena: 😱 I didn't know that 😄

PR #388, opened 2020-05-19
Add support for postgres client. 

mzfr:
I have also formatted some code in the server.py file.

@afeena @glaslos Can you please take a look at the table schema and let me know if any changes is needed in this. If not then I'll continue the work.


tanner/postgres_client.py
```diff
@@ -0,0 +1,94 @@
...
+                        "location.zip_code" INT NULL,
+                        "user_agent" TEXT NOT NULL,
+                        "start_time" TIMESTAMP,
+                        "end_time" TIMESTAMP,
+                        "requests_in_second" FLOAT NULL,
+                        "approx_time_between_requests" FLOAT NULL,
+                        "accepted_paths" INT NULL,
+                        "errors" INT NULL,
+                        "hidden_links" INT NULL,
+                        "attack_count.index" INT NULL,
```
afeena, line 77: should not be here: 1) we have more types than index 2) we can count using paths table

tanner/postgres_client.py
```diff
@@ -0,0 +1,94 @@
...
+                        "paths" TEXT REFERENCES "Paths"(sess_uuid),
+                        "cookies" TEXT REFERENCES "Cookies"(sess_uuid),
+                        "referer" TEXT NULL,
+                        "possible_owners.user" INT NULL
```
afeena, line 81: What about other type? tool? crawler? and it's probability - so definitely not int

tanner/postgres_client.py
```diff
@@ -0,0 +1,94 @@
...
+            async with pg_client.acquire() as conn:
+                async with conn.cursor() as cur:
+                    await cur.execute(
+                        """
+                    CREATE TABLE IF NOT EXISTS "Paths" (
+                        "sess_uuid" TEXT PRIMARY KEY NOT NULL UNIQUE,
+                        "path" TEXT NOT NULL,
+                        "timestamp" FLOAT NOT NULL,
+                        "response_status" INT NOT NULL,
+                        "attack_type" TEXT NOT NULL
```
afeena, line 43: we might change type on int, use enum here

tanner/postgres_client.py
```diff
@@ -0,0 +1,94 @@
...
+            )
+
+            async with pg_client.acquire() as conn:
+                async with conn.cursor() as cur:
+                    await cur.execute(
+                        """
+                    CREATE TABLE IF NOT EXISTS "Paths" (
+                        "sess_uuid" TEXT PRIMARY KEY NOT NULL UNIQUE,
+                        "path" TEXT NOT NULL,
+                        "timestamp" FLOAT NOT NULL,
```
afeena, line 41: timestamp?

tanner/postgres_client.py
```diff
@@ -0,0 +1,94 @@
...
+        try:
+            dsn = "dbname={} user={} password={} host={} port={}".format(
+                self.db_name, self.user, self.password, self.host, self.port
+            )
+
+            pg_client = await asyncio.wait_for(
+                aiopg.create_pool(dsn, maxsize=self.poolsize), timeout=int(self.timeout)
+            )
+
+            async with pg_client.acquire() as conn:
```
afeena, line 34: For me looks like mixing of connection creation and table routine

tanner/postgres_client.py
```diff
@@ -0,0 +1,94 @@
...
+                        "response_status" INT NOT NULL,
+                        "attack_type" TEXT NOT NULL
+                    )
+                    """
+                    )
+
+                    await cur.execute(
+                        """
+                    CREATE TABLE IF NOT EXISTS "Cookies" (
+                        "sess_uuid" TEXT PRIMARY KEY NOT NULL UNIQUE,
```
afeena, line 51: https://www.postgresql.org/docs/9.1/datatype-uuid.html

tanner/postgres_client.py
```diff
@@ -0,0 +1,94 @@
...
+                        "timestamp" FLOAT NOT NULL,
+                        "response_status" INT NOT NULL,
+                        "attack_type" TEXT NOT NULL
+                    )
+                    """
+                    )
+
+                    await cur.execute(
+                        """
+                    CREATE TABLE IF NOT EXISTS "Cookies" (
```
afeena, line 50: please make the table's name either all with first capital letter or all lowercase :D

tanner/postgres_client.py
```diff
@@ -0,0 +1,94 @@
...
+                        "key" TEXT NULL,
+                        "value" TEXT NULL
+                    )
+                    """
+                    )
+
+                    await cur.execute(
+                        """
+                    CREATE TABLE IF NOT EXISTS "session_data" (
+                        "sess_uuid" TEXT NOT NULL PRIMARY KEY,
```
afeena, line 61: same here, I think it's better to use uuid type

tanner/postgres_client.py
```diff
@@ -0,0 +1,94 @@
...
+                        "snare_uuid" TEXT NOT NULL,
+                        "peer.ip" TEXT NOT NULL,
```
afeena, line 63: https://www.postgresql.org/docs/9.1/datatype-net-types.html

tanner/postgres_client.py
mzfr: In paths we have a timestamp 

image

tanner/postgres_client.py
mzfr: so should I remove this column?

tanner/postgres_client.py
mzfr: Yes that can be made float but how do I decide what all columns are needed? I mean in some sessions there is only `user` and in some it's `user` and `tool`.

tanner/postgres_client.py
mzfr: This was the way recommended in the aiopg docs. So should I make a new function called `create_tables` and pass the `pg_client` object to that?

tanner/postgres_client.py
afeena: I think you can do it

tanner/postgres_client.py
afeena: We can store all columns, we have finite set of possible owners, so some of them might be just 0

tanner/postgres_client.py
afeena: @mzfr I meant the type of timestamp :)

mzfr: oh :facepalm:

tanner/postgres_client.py
mzfr: ok so I'll just make the columns and they will be nullable

tanner/postgres_client.py
```diff
@@ -0,0 +1,96 @@
...
+                        "value" TEXT NULL
+                    )
+                    """
+                    )
+
+                    await cur.execute(
+                        """
+                    CREATE TABLE IF NOT EXISTS "session_data" (
+                        "sess_uuid" UUID NOT NULL PRIMARY KEY,
+                        "snare_uuid" TEXT NOT NULL,
```
afeena, line 62: maybe it's better to have it also as UUID type?

afeena:
@glaslos We want to use both redis and pg in tanner: redis as a cache for sessions before analysis (since we delete them afterwards, redis indeed the cache) and pg for storing analysed sessions. In such a case we might separate analysis process into separate background task. 
Also we thought about keeping dorks in redis, because for each request we need access to them, and it would be faster.

What do you think?

glaslos:
This makes a lot of sense. You could even use a TTL to close, push to pg and then delete sessions. You can subscribe to TTL events.

tanner/create_tables.py
```diff
@@ -0,0 +1,71 @@
...
+
+        Arguments:
+            pg_client {aiopg.pool.Pool}
+        """
+        async with pg_client.acquire() as conn:
+            async with conn.cursor() as cur:
+                await cur.execute(
+                    """
+                CREATE TABLE IF NOT EXISTS "paths" (
+                    "sess_uuid" UUID PRIMARY KEY NOT NULL UNIQUE,
```
glaslos, line 18: This shouldn't be a primary key but a foreign key constraint to the `sess_uuid` in the `session_data` table.
Besides that, a primary key is always not null and unique.

tanner/create_tables.py
```diff
@@ -0,0 +1,71 @@
...
+                    "path" TEXT NOT NULL,
+                    "timestamp" TIMESTAMP NOT NULL,
```
glaslos, line 20: Here you are using a name which duplicates the type, use the opportunity to describe the purpose of this field. I assume something like `create_at` maybe? You can also give it a default value like `now()` so you don't have to create it client side and let the database populate it for you.

tanner/create_tables.py
```diff
@@ -0,0 +1,71 @@
...
+                    "response_status" INT NOT NULL,
+                    "attack_type" INT NOT NULL
```
glaslos, line 22: Create an additional table with attack types. Integer id and a string with description (e.g. sqli, xss...) and the add a foreign key constraint here.

tanner/create_tables.py
```diff
@@ -0,0 +1,71 @@
...
+                )
+                """
+                )
+
+                await cur.execute(
+                    """
+                CREATE TABLE IF NOT EXISTS "cookies" (
+                    "sess_uuid" UUID PRIMARY KEY NOT NULL UNIQUE,
```
glaslos, line 30: foreign key constraint

tanner/create_tables.py
```diff
@@ -0,0 +1,71 @@
...
+                    "timestamp" TIMESTAMP NOT NULL,
+                    "response_status" INT NOT NULL,
+                    "attack_type" INT NOT NULL
+                )
+                """
+                )
+
+                await cur.execute(
+                    """
+                CREATE TABLE IF NOT EXISTS "cookies" (
```
glaslos, line 29: don't you have multiple key, value pairs per session? Your `sess_uuid` can't be unique and primary then.

tanner/create_tables.py
```diff
@@ -0,0 +1,71 @@
...
+                    "sess_uuid" UUID PRIMARY KEY NOT NULL UNIQUE,
+                    "key" TEXT NULL,
+                    "value" TEXT NULL
+                )
+                """
+                )
+
+                await cur.execute(
+                    """
+                CREATE TABLE IF NOT EXISTS "session_data" (
```
glaslos, line 39: How about calling this `sessions`?

tanner/create_tables.py
```diff
@@ -0,0 +1,71 @@
...
+                    "sess_uuid" UUID NOT NULL PRIMARY KEY,
```
glaslos, line 40: This looks like `tablename_type` naming to me. Pick something which isn't already there. This field is used to identify a session. Typically we would call this field `id`. `NOT NULL` is given for the primary key.

tanner/create_tables.py
```diff
@@ -0,0 +1,71 @@
...
+                    "snare_uuid" TEXT NOT NULL,
```
glaslos, line 41: call this `sensor_id`. Call me an optimist, but we might have other sensors in the future using Tanner as a service ;)

tanner/create_tables.py
```diff
@@ -0,0 +1,71 @@
...
+                    "peer.ip" INET NOT NULL,
```
glaslos, line 42: use `source` and `destination` to identify an IP address origin.

tanner/create_tables.py
```diff
@@ -0,0 +1,71 @@
...
+                    "peer.port" INT NOT NULL,
+                    "location.country" TEXT NULL,
+                    "location.country_code" TEXT NULL,
+                    "location.city" TEXT NULL,
+                    "location.zip_code" INT NULL,
+                    "user_agent" TEXT NOT NULL,
+                    "start_time" TIMESTAMP,
```
glaslos, line 49: use `now()` as default. start and end should not be null.

tanner/create_tables.py
```diff
@@ -0,0 +1,71 @@
...
+                )
+
+                await cur.execute(
+                    """
+                CREATE TABLE IF NOT EXISTS "session_data" (
+                    "sess_uuid" UUID NOT NULL PRIMARY KEY,
+                    "snare_uuid" TEXT NOT NULL,
+                    "peer.ip" INET NOT NULL,
+                    "peer.port" INT NOT NULL,
+                    "location.country" TEXT NULL,
```
glaslos, line 44: @afeena how about putting location into a separate table?

tanner/create_tables.py
```diff
@@ -0,0 +1,71 @@
...
+                    "location.country_code" TEXT NULL,
+                    "location.city" TEXT NULL,
+                    "location.zip_code" INT NULL,
+                    "user_agent" TEXT NOT NULL,
+                    "start_time" TIMESTAMP,
+                    "end_time" TIMESTAMP,
+                    "requests_in_second" FLOAT NULL,
```
glaslos, line 51:
```suggestion
                    "requests_per_second" FLOAT NULL,
```
you could also abbreviate it to `rps` and add a comment explaining the field: `comment on column rps is 'requests per second';`

tanner/create_tables.py
```diff
@@ -0,0 +1,71 @@
...
+                    "approx_time_between_requests" FLOAT NULL,
+                    "accepted_paths" INT NULL,
+                    "errors" INT NULL,
+                    "hidden_links" INT NULL,
+                    "paths" UUID REFERENCES "paths"(sess_uuid),
```
glaslos, line 56: you want to make the reference the other way around. From the auxiliary tables to the main table using the session id.

tanner/create_tables.py
```diff
@@ -0,0 +1,71 @@
...
+                    "cookies" UUID REFERENCES "cookies"(sess_uuid),
+                    "referer" TEXT NULL,
+                    "possible_owners.user" FLOAT NULL,
+                    "possible_owners.type" FLOAT NULL,
+                    "possible_owners.tool" FLOAT NULL,
+                    "possible_owners.crawler" FLOAT NULL
+                    )
+                """
+                )
+                await cur.execute("CREATE INDEX ON session_data(sess_uuid)")
```
glaslos, line 66: primary key has an index by default.

tanner/create_tables.py
```diff
@@ -0,0 +1,71 @@
...
+                    "end_time" TIMESTAMP,
+                    "requests_in_second" FLOAT NULL,
+                    "approx_time_between_requests" FLOAT NULL,
+                    "accepted_paths" INT NULL,
+                    "errors" INT NULL,
+                    "hidden_links" INT NULL,
+                    "paths" UUID REFERENCES "paths"(sess_uuid),
+                    "cookies" UUID REFERENCES "cookies"(sess_uuid),
+                    "referer" TEXT NULL,
+                    "possible_owners.user" FLOAT NULL,
```
glaslos, line 59: this should be a separate table IMO. you don't want to add a new column whenever we add a new owner type.

tanner/postgres_client.py
```diff
@@ -0,0 +1,36 @@
...
+        pg_client = None
+        try:
+            dsn = "dbname={} user={} password={} host={} port={}".format(
+                self.db_name, self.user, self.password, self.host, self.port
+            )
+
+            pg_client = await asyncio.wait_for(
+                aiopg.create_pool(dsn, maxsize=self.poolsize), timeout=int(self.timeout)
+            )
+        except asyncio.TimeoutError as timeout_error:
```
glaslos, line 33: what about other errors?

tanner/create_tables.py
mzfr: @afeena  suggested that we can have a dictionary which resolves the attack type to an integer something like:
```json
{
	"sqli":1,
	"xss":2,
}
```
And in DB we can store the integer, since it will be cheaper to store.

tanner/create_tables.py
mzfr: if we use NOW() for both start and end then do we need to provide the `NOT NULL`? Asking since now would always provide the value so it won't be null right?

tanner/create_tables.py
glaslos: yeah, same suggestions except in mine the dictionary will be in a table as well. On startup you can fetch that table and populate your in memory dictionary.

tanner/create_tables.py
glaslos: Correct, with DEFAULT you don't need NOT NULL except if you want to prevent the client from explicitly setting it to null.
Not sure if we want to implicitly set the end time by using `DEFAULT now()` or have it NULL until we explicitly end the session. What do you think?

mzfr: I don't see a problem in any of the cases.

mzfr:
@glaslos I've made the requested changes. Can you please review the table schema now?

tanner/create_tables.py
```diff
@@ -0,0 +1,69 @@
...
+            pg_client {aiopg.pool.Pool}
+        """
+        async with pg_client.acquire() as conn:
+            async with conn.cursor() as cur:
+
+                await cur.execute(
+                    """
+                CREATE TABLE IF NOT EXISTS "sessions" (
+                    "id" UUID PRIMARY KEY,
+                    "sensor_id" TEXT NOT NULL,
```
glaslos, line 20: This should be a UUID as well

tanner/create_tables.py
```diff
@@ -0,0 +1,69 @@
...
+                    "possible_owners.type" FLOAT NULL,
+                    "possible_owners.tool" FLOAT NULL,
+                    "possible_owners.crawler" FLOAT NULL
+                    )
+                """
+                )
+                await cur.execute(
+                    """
+                CREATE TABLE IF NOT EXISTS "paths" (
+                    "sess_uuid" UUID REFERENCES sessions(id),
```
glaslos, line 46:
```suggestion
                    "session_id" UUID REFERENCES sessions(id),
```

tanner/create_tables.py
```diff
@@ -0,0 +1,69 @@
...
+                    "path" TEXT NOT NULL,
+                    "created_at" TIMESTAMP NOT NULL,
```
glaslos, line 48: default now()

tanner/create_tables.py
```diff
@@ -0,0 +1,69 @@
...
+                    "response_status" INT NOT NULL,
+                    "attack_type" INT NOT NULL
+                )
+                """
+                )
+
+                await cur.execute(
+                    """
+                CREATE TABLE IF NOT EXISTS "cookies" (
+                    "sess_uuid" UUID REFERENCES sessions(id),
```
glaslos, line 58:
```suggestion
                    "session_id" UUID REFERENCES sessions(id),
```

tanner/create_tables.py
```diff
@@ -0,0 +1,69 @@
...
+                    "key" TEXT NULL,
+                    "value" TEXT NULL
+                )
+                """
+                )
+                await cur.execute("comment on column sessions.rps is 'requests per second'")
+                await cur.execute("CREATE INDEX ON sessions(sensor_id)")
```
glaslos, line 65: why are you adding those indexes?

mzfr: to speed up the process of searching. Like if we want to take out all the sessions having a single snare_id then we can do that.

tanner/create_tables.py
afeena: I thought about this table in the beginning when I started to think about schema. But then I realise that we create attack type table from tanner and convert types inside tanner to put in into the table and then we fetch it from table... For me, it's a little bit redundant.

I prefer to have the Enum class in tanner which represents attack types and we can easily do conversion inside tanner. But if you think it would be better to have a table, I have no complains

afeena: From the other hand, it would be easier to fetch only attack types if we want to do analysis only on that level, so in this case I agree that separate table is much more convenient

glaslos: No, sounds good to me. Adding that table later would be easy 👍

tanner/create_tables.py
afeena: @glaslos That's a good point!

tanner/create_tables.py
afeena: @glaslos Sure, if on the stage of requests implementation it would be easier to have separate table, we always can create it :)

tanner/create_tables.py
afeena: @glaslos I just worry that it might slow down the requests: if we store like this, instead of joining the tables, when we want to filter sessions by location, we just use simple `where`.
I don't think we need location info without session, what do you think?

glaslos: fine by me. I usually don't care too much about location information. We can still refactor once we feel like the sessions table is too busy

tanner/sessions/session_manager.py
```diff
@@ -66,21 +66,21 @@ def get_session_id(self, data):
 
         return hashlib.md5(sess_id_string.encode()).hexdigest()
 
-    async def delete_old_sessions(self, redis_client):
+    async def delete_old_sessions(self, redis_client, pg_client):
         id_for_deletion = [sess_id for sess_id, sess in self.sessions.items() if sess.is_expired()]
         for sess_id in id_for_deletion:
-            is_deleted = await self.delete_session(self.sessions[sess_id], redis_client)
+            is_deleted = await self.delete_session(self.sessions[sess_id], redis_client, pg_client)
```
afeena, line 72: if we agree that we use redis to store intermediate sessions, why we need postgres client here?

tanner/sessions/session_analyzer.py
```diff
@@ -8,6 +8,8 @@
 from tanner.dorks_manager import DorksManager
 from tanner.config import TannerConfig
 
+COOKIE_INSERT_QUERY = "INSERT INTO cookies(sess_uuid, key, value) VALUES({uuid}, {k}, {v})"
```
afeena, line 11: why it's here?

tanner/sessions/session_manager.py
mzfr: I think because there are lines in the `delete_session` function like:

```python
try:
await redis_client.set(sess.get_uuid(), sess.to_json())
await self.analyzer.analyze(sess.get_uuid(), redis_client, pg_client)
```
so I think it analyze them first and then in `save_sessions` function of `session_analyzer` we have deletion of the said key.

tanner/sessions/session_analyzer.py
mzfr: oh this is just while testing. I'll remove the query or will use sqlalchemy for the execution of the query

afeena: Ok :) please use something like TODO to highlight the parts which are still work in progress :)

tanner/sessions/session_manager.py
afeena: My bad!

afeena: But I think we can rewrite this part and make analyzer as a separate background task, to avoid using to DB connections in one place

tanner/create_tables.py
```diff
@@ -0,0 +1,76 @@
...
+                    "key" TEXT NULL,
+                    "value" TEXT NULL
+                )
+                """
+                )
+                await cur.execute(
+                    """
+                CREATE TABLE IF NOT EXISTS "owners" (
+                    "session_id" UUID REFERENCES sessions(id),
+                    "key" TEXT,
```
afeena, line 64: please use more specific column names, smth like "owner_type"/"probability"

tanner/server.py
```diff
@@ -126,20 +128,28 @@ def create_app(self, loop):
...
+        app["session_delete"].cancel()
+        await app["session_delete"]
 
     def start(self):
         loop = asyncio.get_event_loop()
-        self.redis_client = loop.run_until_complete(redis_client.RedisClient.get_redis_client())
+        self.redis_client = loop.run_until_complete(
+            redis_client.RedisClient.get_redis_client()
+        )
+        self.pg_client = loop.run_until_complete(
```
afeena, line 142: Maybe we can do it like aiohttp suggests? https://docs.aiohttp.org/en/stable/faq.html#where-do-i-put-my-database-connection-so-handlers-can-access-it

tanner/sessions/session_analyzer.py
```diff
@@ -27,18 +33,132 @@ async def analyze(self, session_key, redis_client):
         else:
             result = await self.create_stats(session, redis_client)
             await self.queue.put(result)
-            await self.save_session(redis_client)
+            await self.save_session(redis_client, pg_client)
 
-    async def save_session(self, redis_client):
+    async def save_session(self, redis_client, pg_client):
         while not self.queue.empty():
+            columns = (
```
afeena, line 40: you have a class for schema creation, you can make it more general and fetch the columns from it

tanner/sessions/session_analyzer.py
```diff
@@ -27,18 +33,132 @@ async def analyze(self, session_key, redis_client):
...
+                        et=end_time,
+                        rps=session["requests_in_second"],
+                        atbr=session["approx_time_between_requests"],
+                        apaths=session["accepted_paths"],
+                        err=session["errors"],
+                        hlinks=session["hidden_links"],
+                        referer=session["referer"]
+                    )
+                else:
+                    all_columns = columns.format("country")
```
afeena, line 97: I don't like such code duplication. Try to make only one request and put the data into unified format, you can use `get` on location and set default value if some data is not available

tanner/sessions/session_analyzer.py
```diff
@@ -27,18 +33,132 @@ async def analyze(self, session_key, redis_client):
...
+                    sessions_query = (
+                        "INSERT INTO sessions ({cols}) "
+                        "VALUES ('{uuid}','{sensor}','{ip}',{port},'{country}',"
+                        "'{ua}','{st}','{et}',{rps},{atbr},{apaths},{err},{hlinks},"
+                        "'{referer}');"
+                    ).format(
```
afeena, line 103: I think using placeholders will be better than using format
https://aiopg.readthedocs.io/en/stable/core.html#aiopg.Cursor.execute
https://www.psycopg.org/docs/usage.html#query-parameters

mzfr: Placeholders like `%s`  or like `%(id)s` this?

tanner/sessions/session_analyzer.py
```diff
@@ -2,12 +2,18 @@
...
 from geoip2.database import Reader
 import geoip2
 import aioredis
 from tanner.dorks_manager import DorksManager
 from tanner.config import TannerConfig
 
+# TODO: Move Query from here
+COOKIE_INSERT_QUERY = "INSERT INTO cookies(session_id, key, value) VALUES('{uuid}', '{key}', '{value}');"
+# TODO: Better way to resolve the value to int
+ATTACK_TYPE = {"index": 1, "sqli": 2, "rfi": 3, "lfi": 4}
```
afeena, line 15: https://docs.python.org/3.7/library/enum.html

tanner/sessions/session_analyzer.py
```diff
@@ -27,18 +31,76 @@ async def analyze(self, session_key, redis_client):
...
+            s_key = session["snare_uuid"]
+            del_key = session["sess_uuid"]
+            print("Printing sessions")
+
+            start_time = datetime.fromtimestamp(session["start_time"]).strftime('%Y-%m-%d %H:%M:%S')
+            end_time = datetime.fromtimestamp(session["end_time"]).strftime('%Y-%m-%d %H:%M:%S')
+
             try:
-                await redis_client.zadd(s_key, session['start_time'], json.dumps(session))
+                sessions_query = self.insert_query.Sessions.format(
```
afeena, line 52: you should not build the queries inside analysis. You have a class, use this class to incapsulate all the operations with DB. Ideally, analysis class should not know ANYTHING about db structure and queries

tanner/dbutils.py
```diff
@@ -0,0 +1,100 @@
+import asyncio
+
+
+class CreateTables:
```
afeena, line 4: please differentiate between class and function. I prefer to have a class as a Noun entity which consists of actions: 
e.g. 
```
class DBUtils <Noun>:
<...>
    def create_tables <action (verb)>:
        <...>

```

tanner/dbutils.py
```diff
@@ -0,0 +1,100 @@
...
+                )
+                await cur.execute("CREATE INDEX ON sessions(sensor_id)")
+                await cur.execute('CREATE INDEX ON "paths"(session_id)')
+                await cur.execute('CREATE INDEX ON "cookies"(session_id)')
+                await cur.execute('CREATE INDEX ON "owners"(session_id)')
+            cur.close()
+        conn.close()
+
+
+class InsertionQueries:
```
afeena, line 83: I suggest you to collect everything which related to our tanner DB inside one class. What is the purpose of the separate class? And why you have only constants here instead of doing queries?

tanner/sessions/session_analyzer.py
```diff
@@ -27,18 +31,76 @@ async def analyze(self, session_key, redis_client):
...
+                            await cur.execute(paths_query)
+
+                        for k, v in session["possible_owners"].items():
+                            await cur.execute(self.insert_query.Owners.format(id=session["sess_uuid"], key=k, val=v))
+
+                    cur.close()
+                conn.close()
+
                 await redis_client.delete(*[del_key])
             except aioredis.ProtocolError as redis_error:
```
afeena, line 99: what about aiopg exceptions?
https://aiopg.readthedocs.io/en/stable/core.html#exceptions

tanner/sessions/session_analyzer.py
```diff
@@ -27,18 +31,76 @@ async def analyze(self, session_key, redis_client):
...
+                    et=end_time,
+                    rps=session["requests_in_second"],
+                    atbr=session["approx_time_between_requests"],
+                    apaths=session["accepted_paths"],
+                    err=session["errors"],
+                    hlinks=session["hidden_links"],
+                    referer=session["referer"]
+                )
+
+                async with pg_client.acquire() as conn:
```
afeena, line 72: same here. Please move all the queries to DB class.

tanner/postgres_client.py
afeena: @mzfr Please take a look https://aiopg.readthedocs.io/en/stable/core.html#exceptions :)

tanner/dbutils.py
```diff
@@ -0,0 +1,183 @@
...
+                    "value" TEXT NULL
+                )
+                """
+                )
+                await cur.execute(
+                    """
+                CREATE TABLE IF NOT EXISTS "owners" (
+                    "session_id" UUID REFERENCES sessions(id),
+                    "owner_type" TEXT,
+                    "probability" TEXT
```
afeena, line 70: maybe float is better?

tanner/dbutils.py
```diff
@@ -0,0 +1,183 @@
...
+                )
+                await cur.execute("CREATE INDEX ON sessions(sensor_id)")
+                await cur.execute('CREATE INDEX ON "paths"(session_id)')
+                await cur.execute('CREATE INDEX ON "cookies"(session_id)')
+                await cur.execute('CREATE INDEX ON "owners"(session_id)')
+            cur.close()
+        conn.close()
+
+    @staticmethod
+    async def insert_queries(session, pg_client):
```
afeena, line 88: maybe we can find some better name for this function? :)

tanner/dbutils.py
```diff
@@ -0,0 +1,183 @@
...
+
+            async with pg_client.acquire() as conn:
+                async with conn.cursor() as cur:
+                    await cur.execute(sessions_query)
+                    print("Inserted sessions")
+                    for k, v in session["cookies"].items():
+                        await cur.execute(
+                            Cookies.format(uuid=session["sess_uuid"], key=k, value=v)
+                        )
+                    print("Inserted Cookies")
```
afeena, line 157: maybe we can log instead of print?

mzfr: oh yes I was testing with these and was going to ask, should we log after every query? or should we just log once  like `Inserted sessions`

because if we log for every query then logs would contain lot of repetitive things because it will be repeated for all the sessions

afeena: @mzfr I think we can remove this and log only exceptions :)

mzfr: I mean if we log only once then we can log with the session UUID so something like: `Inserted sessions with Session UUID <sess_uuid>`

it won't look much repetitive but if we log after every query then a lot of repetition would be there.

But also logging after every query would help us in debugging if something goes wrong :)

so I just wanted to ask your opinion on this.

tanner/server.py
```diff
@@ -126,20 +128,28 @@ def create_app(self, loop):
...
+            postgres_client.PostgresClient().get_pg_client()
```
afeena, line 143: I run the tanner without postgresql running and it doesn't exit with error. Could you please check?

mzfr: You have the Postgres in your config, right? 
I mean I just tested without running Postgres and I got the error 

```
psycopg2.OperationalError: could not connect to server: Connection refused
        Is the server running on host "localhost" (::1) and accepting
        TCP/IP connections on port 5432?
could not connect to server: Connection refused
        Is the server running on host "localhost" (127.0.0.1) and accepting
        TCP/IP connections on port 5432?
```
I did not make any changes to the code or anything I just simply stopped my Postgres and ran `sudo python bin/tanner`

tanner/postgres_client.py
```diff
@@ -0,0 +1,37 @@
...
+        pg_client = None
+        try:
+            dsn = "dbname={} user={} password={} host={} port={}".format(
+                self.db_name, self.user, self.password, self.host, self.port
+            )
+
+            pg_client = await asyncio.wait_for(
+                aiopg.create_pool(dsn, maxsize=self.poolsize), timeout=int(self.timeout)
+            )
+        except (asyncio.TimeoutError, psycopg2.ProgrammingError) as error:
```
afeena, line 34: OK, now I am able to get an exception :) But it's not catched:

```
  File "/usr/local/bin/tanner", line 4, in <module>
    __import__('pkg_resources').run_script('Tanner==0.6.0', 'tanner')
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 661, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/usr/lib/python3.7/site-packages/pkg_resources/__init__.py", line 1448, in run_script
    exec(script_code, namespace, namespace)
  File "/usr/local/lib/python3.7/site-packages/Tanner-0.6.0-py3.7.egg/EGG-INFO/scripts/tanner", line 36, in <module>
  File "/usr/local/lib/python3.7/site-packages/Tanner-0.6.0-py3.7.egg/EGG-INFO/scripts/tanner", line 32, in main
  File "/usr/local/lib/python3.7/site-packages/Tanner-0.6.0-py3.7.egg/tanner/server.py", line 145, in start
  File "uvloop/loop.pyx", line 1456, in uvloop.loop.Loop.run_until_complete
  File "/usr/local/lib/python3.7/site-packages/Tanner-0.6.0-py3.7.egg/tanner/postgres_client.py", line 33, in get_pg_client
  File "/usr/lib64/python3.7/asyncio/tasks.py", line 442, in wait_for
    return fut.result()
  File "/usr/local/lib/python3.7/site-packages/aiopg/utils.py", line 61, in send
    return self._coro.send(value)
  File "/usr/local/lib/python3.7/site-packages/aiopg/pool.py", line 155, in from_pool_fill
    await self._fill_free_pool(False)
  File "/usr/local/lib/python3.7/site-packages/aiopg/pool.py", line 199, in _fill_free_pool
    **self._conn_kwargs)
  File "/usr/local/lib/python3.7/site-packages/aiopg/connection.py", line 551, in _connect
    await self._poll(self._waiter, self._timeout)
  File "/usr/local/lib/python3.7/site-packages/aiopg/connection.py", line 207, in _poll
    await asyncio.wait_for(self._waiter, timeout, loop=self._loop)
  File "/usr/lib64/python3.7/asyncio/tasks.py", line 442, in wait_for
    return fut.result()
  File "/usr/local/lib/python3.7/site-packages/aiopg/connection.py", line 106, in _ready
    state = self._conn.poll()
psycopg2.OperationalError: could not connect to server: Connection refused
	Is the server running on host "localhost" (::1) and accepting
	TCP/IP connections on port 5432?
could not connect to server: Connection refused
	Is the server running on host "localhost" (127.0.0.1) and accepting
	TCP/IP connections on port 5432?
```

tanner/sessions/session_analyzer.py
```diff
@@ -144,11 +144,13 @@ def find_location(ip):
...
         try:
             location = reader.city(ip)
+            if location.postal.code is None:
+                zcode = 0
             info = dict(
                 country=location.country.name,
                 country_code=location.country.iso_code,
                 city=location.city.name,
-                zip_code=location.postal.code,
+                zip_code=zcode,
```
afeena, line 153: and what if the `location.postal.code` is not None? :)

tanner/server.py
```diff
@@ -101,23 +103,23 @@ async def handle_version(self, request):
         return web.json_response(response_msg)
 
     async def on_shutdown(self, app):
-        await self.session_manager.delete_sessions_on_shutdown(self.redis_client)
+        await self.session_manager.delete_sessions_on_shutdown(self.redis_client, self.pg_client)
         self.redis_client.close()
         await self.redis_client.wait_closed()
```
afeena, line 108: please close also aiopg pool

tanner/sessions/session_analyzer.py
mzfr: If it's none then it would be an integer right?

mzfr: oh wait, those are numbers but still passed as string. I'll do the conversion.

tanner/server.py
```diff
@@ -106,6 +106,7 @@ async def on_shutdown(self, app):
         await self.session_manager.delete_sessions_on_shutdown(self.redis_client, self.pg_client)
         self.redis_client.close()
         await self.redis_client.wait_closed()
+        await self.pg_client.wait_closed()
```
afeena, line 109: why you call only `wait_closed` without closing the pool? You closed it somewhere else?
https://aiopg.readthedocs.io/en/stable/core.html#aiopg.Pool.close

tanner/sessions/session_analyzer.py
afeena: `UnboundLocalError: local variable 'zcode' referenced before assignment`

tanner/sessions/session_analyzer.py
```diff
@@ -138,14 +144,22 @@ def find_location(ip):
...
         try:
             location = reader.city(ip)
+            if location.postal.code is None:
+                zcode = 0
             info = dict(
                 country=location.country.name,
                 country_code=location.country.iso_code,
                 city=location.city.name,
-                zip_code=location.postal.code,
+                zip_code=int(zcode),
```
afeena, line 153: `UnboundLocalError: local variable 'zcode' referenced before assignment`

tanner/server.py
mzfr: Sorry I missed that :sweat_smile:

afeena:
Great job!

Things to be considered for the future:
-- Splitting the analysis part and core tanner part
-- Tests

PR #390, opened 2020-06-04
Deploy master branch to server using github actions

mzfr:
@glaslos Can you please review this and let me know what all changes are required in this?


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,43 @@
...
+  push:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
```
glaslos, line 13: after checking out the code, build the .egg or .wheel

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,43 @@
...
+    - name: Copy file via scp
+      uses: appleboy/scp-action@master
+      env:
+        HOST: ${{ secrets.HOST }}
+        USERNAME: ${{ secrets.USERNAME }}
+        PORT: ${{ secrets.PORT }}
+        KEY: ${{ secrets.SSHKEY }}
+      with:
+        source: "."
```
glaslos, line 22: only scp the .egg

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,43 @@
...
+    - name: Install dependencies
+      uses: appleboy/ssh-action@master
+      with:
+        host: ${{ secrets.HOST }}
+        USERNAME: ${{ secrets.USERNAME }}
+        PORT: ${{ secrets.PORT }}
+        KEY: ${{ secrets.SSHKEY }}
+        script: |
+        python -m pip install --upgrade pip
+        pip3 install -r requirements.txt
```
glaslos, line 34: is there any other way to do this?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,43 @@
...
+      
+    - name: Starting tanner
+      uses: appleboy/ssh-action@master
+      with:
+        host: ${{ secrets.HOST }}
+        USERNAME: ${{ secrets.USERNAME }}
+        PORT: ${{ secrets.PORT }}
+        KEY: ${{ secrets.SSHKEY }}
+        script: supervisorctl restart service
```
glaslos, line 43: can you skip this step and have tanner auto restart if there is a new egg in a folder?

.github/workflows/deploy.yml
mzfr: another way to install the dependencies? That is what was suggested in the github action documentation.

glaslos: Is it the only way? Do some reseach

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,35 @@
...
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.6
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.6
```
afeena, line 17: why we build it with py3.6 if we do not support it anymore?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,35 @@
...
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+      
+    - name: Generate .egg
```
afeena, line 23: what about tests?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,35 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.6
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.6
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
```
afeena, line 21: we rarely change dependencies, maybe it makes sense to add cache?  https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,35 @@
...
+    - name: Copy file via scp
+      uses: appleboy/scp-action@master
+      env:
+        HOST: ${{ secrets.HOST }}
+        USERNAME: ${{ secrets.USERNAME }}
+        PORT: ${{ secrets.PORT }}
+        KEY: ${{ secrets.SSHKEY }}
+      with:
+        source: "/dist/*.egg"
+        target: "/opt/tanner/dist"
```
afeena, line 35: I didn't get how tanner suppose to restart here

PR #392, opened 2020-06-08
Update existing tests

mzfr:
In this PR I've updated all the existing tests that were failing because of all the new changes made to the code.

### DIFF
--- a/docker/postgres/dockerfile
+++ b/docker/postgres/dockerfile
@@ -1,4 +1,4 @@
-FROM redis:alpine
+FROM postgres:alpine
 
 # Setup apt
 RUN sed -i 's/dl-cdn/dl-2/g' /etc/apk/repositories && \
--- a/docker/postgres/dockerfile
+++ b/docker/postgres/dockerfile
@@ -1,0 +1,13 @@
+FROM redis:alpine
+
+# Setup apt
+RUN sed -i 's/dl-cdn/dl-2/g' /etc/apk/repositories && \
+    apk -U --no-cache add postgresql && \
+    apk -U --no-cache add postgresql-contrib && \
+# Clean up
+    rm -rf /tmp/* /var/tmp/* && \
+    rm -rf /var/cache/apk/*
+
+# Start redis
+USER nobody:nobody
+CMD systemctl start postgres
--- a/tanner/tests/test_api.py
+++ b/tanner/tests/test_api.py
@@ -123,7 +123,7 @@ class TestApi(unittest.TestCase):
             self.result = await self.handler.return_snare_info(snare_id, count, offset)
 
         self.loop.run_until_complete(test(SESSION_DATA["snare_uuid"]))
-        self.assertEquals(len(self.result), 1)
+        self.assertEqual(len(self.result), 1)
         self.assertEqual(SESSION_DATA["sess_uuid"], self.result[0]["id"])
         self.assertEqual(
             SESSION_DATA["accepted_paths"], self.result[0]["accepted_paths"]
@@ -145,7 +145,6 @@ class TestApi(unittest.TestCase):
 
         async def test(snare_id):
             self.result = await self.handler.return_snare_info(snare_id, count, offset)
-            print(self.result)
 
         self.loop.run_until_complete(test("9f7d7dd3-ac6b-468b-8cee-"))
         self.assertIn("Invalid SNARE UUID", self.result)
@@ -256,7 +255,7 @@ class TestApi(unittest.TestCase):
         )
 
         self.returned_content = self.handler.apply_filters(filters)
-        self.assertEquals(self.returned_content, self.expected_content)
+        self.assertEqual(self.returned_content, self.expected_content)
 
     def test_apply_filter_end_time(self):
         filters = {
@@ -271,7 +270,7 @@ class TestApi(unittest.TestCase):
         )
 
         self.returned_content = self.handler.apply_filters(filters)
-        self.assertEquals(self.returned_content, self.expected_content)
+        self.assertEqual(self.returned_content, self.expected_content)
 
     def tearDown(self):
         async def close():
--- a/tanner/tests/test_api.py
+++ b/tanner/tests/test_api.py
@@ -300,4 +300,5 @@ class TestApi(unittest.TestCase):
 
     @pytest.fixture(scope="session")
     async def close_connection(self):
-        await self.pg_client.close()
+        self.pg_client.close()
+        await self.pg_client.wait_closed()
--- a/tanner/tests/test_api.py
+++ b/tanner/tests/test_api.py
@@ -4,6 +4,7 @@ import aioredis
 import itertools
 import sqlalchemy
 from unittest import mock
+from aiopg.sa import create_engine
 from tanner.api.api import Api
 from tanner import postgres_client
 from tanner.dbutils import DBUtils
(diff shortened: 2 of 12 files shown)


afeena:
Could you please take a look why the tests build has error state?

mzfr:
Yes I had taken a look and it seems that there is some issue in installing the `aiopg`package. Not Sure why that is happening. I'll try to find the solution for this ASAP.

tanner/tests/test_dbutils.py
```diff
@@ -0,0 +1,29 @@
+import asyncio
+import asynctest
```
afeena, line 2: do you use it? either add it to requirements or delete from here :)

mzfr: Yeah, I'll remove them since asynctest wasn't able to help much in mocking Postgres.

afeena:
@mzfr please fix the travis build, seems like some tests are broken

mzfr:
@afeena Tests seems to work locally but they are failing due to postgres not found on `127.0.0.1`. Not sure why the docker file is not getting build.

afeena:
@mzfr Have you defined postgres in travis config?

docker/postgres/dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM redis:alpine
```
afeena, line 1: why are you using redis base image?

tanner/tests/test_api.py
```diff
@@ -5,345 +5,272 @@
...
         self.expected_content = None
         self.conn = None
         self.key = None
 
         async def connect():
-            self.redis_client = await redis_client.RedisClient.get_redis_client()
+            self.postgres = postgres_client.PostgresClient()
+            self.postgres.host = "localhost"
+            self.postgres.post = 5432
+            self.postgres.db_name = "postgres"
```
afeena, line 74: I would be super careful of using default db name for tests

afeena: You can create/drop database for tests directly

tanner/tests/test_session_analyzer.py
```diff
@@ -189,6 +192,6 @@ def test_find_location(self):
             country='United States',
```
afeena, line 192: For fail case you can create separate mock which raises and exception instead of returning value

afeena:
@mzfr Could you please take a look into the warning as well `/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/aiopg/pool.py:310: ResourceWarning: Unclosed 1 connections in <aiopg.pool.Pool object at 0x7fb8ebf3add8>
  ResourceWarning)`

docker/postgres/dockerfile
afeena: @mzfr

tanner/tests/test_api_server.py
```diff
@@ -88,21 +88,21 @@ async def mock_return_snare_stats(snare_uuid):
...
         request = await self.client.request(
             "GET",
-            "/8fa6aa98-4283-4085-bfb9-a1cd3a9e56e4/sessions?filters=peer_ip:127.0.0.1 start_time:1497890400"
+            "/8fa6aa98-4283-4085-bfb9-a1cd3a9e56e4/sessions?filters=peer_ip:127.0.0.1 start_time:11-05-2020"
             " user_agent:ngnix&key=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VyIjoidGFubmVyX293bmVyIn0."
             "NQ7x_iq2t2SUs20Z9G-FmgqeNBOp5duiXr_auNVmzfU"
         )  # noqa
         assert request.status == 200
         detection = await request.json()
+        print(detection)
```
afeena, line 105: I believe we do not need a print

tanner/tests/test_web_server.py
```diff
@@ -9,12 +9,11 @@
 
 class TestWebServer(AioHTTPTestCase):
     def setUp(self):
-        self.loop = asyncio.new_event_loop()
-
         self.handler = TannerWebServer()
-        self.handler.api = AsyncMock()
-        self.handler.redis_client = AsyncMock()
-        self.handler.redis_client.close = AsyncMock()
+        postgres = mock.Mock()
```
afeena, line 13: why don't you use AsyncMock for postgres?

tanner/tests/test_api.py
```diff
@@ -2,348 +2,293 @@
...
 
+        async def connect():
+            self.postgres = postgres_client.PostgresClient()
+            self.postgres.host = "localhost"
+            self.postgres.post = 5432
+            self.postgres.db_name = "tanner_test_db"
+            self.postgres.user = "postgres"
+            self.postgres.password = "postgres"
+            self.postgres.maxsize = 80
+            self.pg_client = await self.postgres.get_pg_client()
```
afeena, line 83: you get the client here and do not close it aftewards

tanner/tests/test_api.py
```diff
@@ -2,348 +2,293 @@
...
-            self.redis_client = await redis_client.RedisClient.get_redis_client()
+        async def create_db():
+            with sqlalchemy.create_engine(
+                "postgresql://postgres@/postgres", isolation_level="AUTOCOMMIT"
+            ).connect() as connection:
+                connection.execute("CREATE DATABASE tanner_test_db")
+                connection.close()
 
+        async def connect():
+            self.postgres = postgres_client.PostgresClient()
```
afeena, line 76: I am just wondering, where this variable is used?

tanner/tests/test_api.py
afeena, line 83: I can't connect locally to the postgres because my password is not the "default"

tanner/tests/test_api.py
```diff
@@ -2,348 +2,293 @@
...
+        self.pg_client = None
         self.returned_content = None
         self.expected_content = None
         self.conn = None
         self.key = None
 
-        async def connect():
-            self.redis_client = await redis_client.RedisClient.get_redis_client()
+        async def create_db():
+            with sqlalchemy.create_engine(
```
afeena, line 69: why don't you use aiopg here?

tanner/tests/test_api.py
mzfr: After this line, we are defining all the variables required to connect to Postgres.
```
        async def connect():
            self.postgres = postgres_client.PostgresClient()
            self.postgres.host = "localhost"
            self.postgres.post = 5432
            self.postgres.db_name = "tanner_test_db"
            self.postgres.user = "postgres"
            self.postgres.password = "postgres"
            self.postgres.maxsize = 80

```

tanner/tests/test_api.py
mzfr: Should I read this information from the config file?

tanner/tests/test_api.py
afeena: 🤦 my bad :)

tanner/tests/test_api_server.py
```diff
@@ -10,8 +11,8 @@ class TestAPIServer(AioHTTPTestCase):
     def setUp(self):
         self.serv = server.ApiServer()
 
-        postgres = mock.Mock()
-        postgres.close = mock.Mock()
+        postgres = AsyncMock()
+        postgres.close = AsyncMock()
```
afeena, line 15: close is not a coroutine https://aiopg.readthedocs.io/en/stable/core.html#aiopg.Pool.close

tanner/tests/test_api.py
```diff
@@ -1,349 +1,303 @@
...
+                        WHERE pg_stat_activity.datname = 'tanner_test_db';
+                    """
+                    )
+                    await conn.execute("DROP database tanner_test_db")
 
         self.loop.run_until_complete(close())
+
+    @pytest.fixture(scope="session")
+    async def close_connection(self):
+        await self.pg_client.close()
```
afeena, line 303: close should not be a coroutine according to docs

afeena: you can also try add https://aiopg.readthedocs.io/en/stable/core.html#aiopg.Pool.wait_closed

PR #396, opened 2020-08-04
update filter search

mzfr:
Now it's possible to search multiple values of the same type.

When I initially made change to the API it only supported filter search in the following way:

```
<url>?filters=attack_type:index
```
but failed to support something like:
```
<url>?filters=owners:tool owners:admin
```

This PR add that support


mzfr:
@afeena what should we do about 3.6 tests failing?

tanner/web/server.py
```diff
@@ -54,15 +54,14 @@ async def handle_sessions(self, request):
...
             if 'filters' in params:
                 for filt in params['filters'].split():
-                    applied_filters[filt.split(':', 1)[0]] = filt.split(':', 1)[1]
-                if 'start_time' in applied_filters:
-                    applied_filters['start_time'] = applied_filters['start_time']
-                if 'end_time' in applied_filters:
-                    applied_filters['end_time'] = applied_filters['end_time']
+                    key, value = filt.split(':', 1)
+                    applied_filters[key].append(value)
+            print(applied_filters)
```
afeena, line 64: remove print

tanner/api/api.py
```diff
@@ -271,33 +270,41 @@ def check_time(time):
...
-            end_time = check_time(filters["end_time"])
-            where += " AND S.end_time<='%s'" % (end_time)
-        if "peer_ip" in filters:
-            where += " AND S.ip='%s'" % (filters["peer_ip"])
-        if "user_agent" in filters:
-            where += " AND S.user_agent='%s'" % (filters["user_agent"])
+        where = "S.sensor_id='%s'" % (filters["sensor_id"][0])
+
+        for parameter, values in filters.items():
+            if parameter == "attack_type":
```
afeena, line 276: I thought a lot about this query building, can we rebuilt it in a way it looks more accurate and we use the power of sqlalchemy? like here https://aiopg.readthedocs.io/en/stable/examples.html#complex-sqlalchemy-queries

mzfr: I actually thought about it and I think almost everything would stay similar to the way it is. 
The only thing that would differ is that instead of having raw queries we would be using the sqlalchemy objects 

I mean all these conditions would have to stay so code might not become as clean as we think.

But I can definitely try to do it.

afeena: We could keep it in mind and make as a separate task :)

afeena:
@mzfr I think we should remove python 3.6 from travis config, what do you think?

PR #399, opened 2020-08-12
Migration script

mzfr:
This script can help people in migrating their data from old redis setup to new postgres model.

### DIFF
--- a/tanner/migration_script.py
+++ b/tanner/migration_script.py
@@ -5,6 +5,16 @@ import psycopg2
 from tanner import redis_client, postgres_client, dbutils
 
 
+async def check_session_data(result):
+    if result["location"] == "NA":
+        result["location"] = dict(
+            country=None, country_code=None, city=None, zip_code=0,
+        )
+
+    if "user_agent" not in result:
+        result["user_agent"] = ""
+
+
 async def main():
     r_client = await redis_client.RedisClient.get_redis_client()
     pg_client = await postgres_client.PostgresClient().get_pg_client()
@@ -22,11 +32,7 @@ async def main():
             try:
                 session = await r_client.zrange(key, encoding="utf-8")
                 result = json.loads(session[0])
-
-                if result["location"] == "NA":
-                    result["location"] = dict(
-                        country=None, country_code=None, city=None, zip_code=0,
-                    )
+                await check_session_data(result)
 
                 try:
                     await dbutils.DBUtils.add_analyzed_data(result, pg_client)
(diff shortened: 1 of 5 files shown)


tanner/migration_script.py
```diff
@@ -0,0 +1,56 @@
...
+                except aioredis.ProtocolError as redis_error:
+                    print(
+                        "Error with redis: %s. Session with session-id %s will not be removed from redis.",
+                        redis_error,
+                        key,
+                    )
+            except aioredis.errors.ReplyError:
+                continue
+
+    self.redis_client.close()
```
afeena, line 49: you have `r_client`, not `self.redis_client`

afeena:
while testing, I got 

```
psycopg2.errors.NotNullViolation: null value in column "user_agent" violates not-null constraint
DETAIL:  Failing row contains (abd0fc36-d96d-4cba-8771-b5417bcc6cb1, 410c054a-3029-4e27-8172-c8679d75634a, 172.19.0.1, 50002, null, null, null, 0, null, 2019-10-12 02:52:00, 2019-10-12 02:52:00, 85598.04081632652, 0, 1, 0, 0, null).
```

afeena:
Don't forget to set values according to its type too :) 

```psycopg2.errors.InvalidTextRepresentation: invalid input syntax for type integer: "N/A"
LINE 1: ...58', 2.8084954716092154, 0.3560633938033859, 231, 'N/A', 'N/... ```

**Comment by afeena** [comment]
@mzfr the problem not only in integer, but in all the types with missing values

```
psycopg2.errors.InvalidTextRepresentation: invalid input syntax for type double precision: "N/A"
LINE 1: ...47:46', '2019-07-26 07:47:46', 89240.51063829787, 'N/A', 1, ...
```

**Comment by mzfr** [comment]
Is it because of the Float and Integer?

File: tanner/migration_script.py [hunk 2]
```diff
@@ -0,0 +1,67 @@
+import asyncio
+import json
+import aioredis
+import psycopg2
+from tanner import redis_client, postgres_client, dbutils
+
+
+async def check_session_data(result):
+    Integers = ["zip_code", "errors", "accepted_paths", "port", "atbr", "rps", "hidden_links"]
```
afeena, line 9: in raw data, there s no atbr and rps,
rps -->requests_in_second
atbr --> approx_time_between_requests

afeena:
In our system,  `zip` code is Integer, but I found out that it's not correct, e.g. in Japan, zip codes contain `-` and some US states have 0 at the beginning of the ZIP code, which is not a valid integer. So I'd recommend changing the type to string

mzfr:
@afeena should I update the code as well as the schema for the postgres?

afeena:
@mzfr yes, please go ahead :)

afeena:
@mzfr I tested it and found some problems of migration, to make it easier, I will merge this PR and upload the changes as a separate commit :)

Issue #381, opened 2020-03-23
GSOC: Adding support for twig template engine


rjt-gupta:
**Aim** - Add support for twig template engine (PHP) for template injection emulator.

**How to do?**

- install twig using composer (https://twig.symfony.com/doc/2.x/installation.html)
- First build a suitable vulnerable code template in which the payload will be injected and then execute using phpox.

The template I had created was - 

```php
<?php 

require '/tmp/vendor/twig/twig/lib/Twig/Autoloader.php';
require '/tmp/vendor/twig/twig/lib/Twig/Extension/StringLoader.php';

Twig_Autoloader::register();
$loader = new Twig_Loader_String();
$twig = new Twig_Environment($loader);
$twig->addExtension(new \Twig\Extension\StringLoaderExtension());
							
 $payload = "<<payload goes here>>";
 $result = $twig->render($payload);
echo $result;
?>
```

This template was working with basic payloads but not fully working with the complex ones (was only working inside phpox not with tanner)

- You can create your own template or can make this one work.

- Test payloads - 
1. Basic payload:  {{3*'2'}}  -> 6    (only twig specific injection)
2. Some complex ones like - {{_self.env.registerUndefinedFilterCallback('shell_exec')}}{{_self.env.getFilter('id')}} or others capable of executing system commands.

- The emulator structure will be same as other ones (php object injection).
- Create a regex also for scanning.
- write tests!

I think this task shouldn't take long with all this information. Feel free to ping us on slack or here for any queries :)

glaslos:
Can you provide some more information on how the request looks like?
Maybe a PCAP from a real, vulnerable system?
Feel free to change the title for something more specific. I think it's a very good proposal for a new vulnerability type.

rjt-gupta:
First i was trying to test the template with major payloads using postman. I also think adding this support would be very cool.

I will try to capture the pcap asap!

PR #401, opened 2020-08-22
Twig template injection 

mzfr:
Add twig template 
Fix for #381

### DIFF
--- a/tanner/emulators/twig_template_injection.py
+++ b/tanner/emulators/twig_template_injection.py
@@ -60,18 +60,7 @@ class TwigTemplateInjection:
             detection = dict(name="twig_template_injection", order=3)
         return detection
 
-    async def handle(self, attack_params):
-        """
-        Handler of emulator
-        :param attack_params (list): contains dicts as elements with id and value (payload from attacker) as keys
-        :return: (dict): value (result of emulator), page (if set to true the payload will be injected to index.html
-        itself) as keys.
-        """
-
-        result = await self.get_injection_result(attack_params[0]["value"])
-        if not result or "stdout" not in result:
-            self.logger.exception(
-                "Error while getting the injection results from php sandbox.."
-            )
-            return dict(status_code=504)
-        return dict(value=result["stdout"], page=False)
+    async def handle(self, attack_params, session=None):
+        attack_params[0]['value'] = unquote(attack_params[0]['value'])
+        result = await self.get_injection_result(attack_params[0]['value'])
+        return result


afeena:
@mzfr And what is the pattern to catch this type of attack?

mzfr:
@afeena This kind of injection can be detected by the same regex used for tornado template injection.

That is why I didn't added any new pattern.

rjt-gupta:
@mzfr Are you able to see the results now?

mzfr:
@rjt-gupta no, still having the same problem i.e no error, no result.

tanner/emulators/twig_template_injection.py
```diff
@@ -0,0 +1,77 @@
...
+    def scan(self, value):
+        """
+        Scans the input payload to detect attack using regex
+        :param value (str): code from attacker
+        :return: detection (dict): name (attack name), order (attack order) as keys
+        """
+
+        detection = None
+        if patterns.TEMPLATE_INJECTION_TORNADO.match(value):
+            detection = dict(name="twig_template_injection", order=3)
```
afeena, line 60: I tested with super simple string /foobar?b={{3*%272%27}} There are 2 possible detections:
```
{'name': 'template_injection', 'order': 4}
{'name': 'twig_template_injection', 'order': 3}
```
Since `template_injection` is order 4, tanner chooses it and execute. 

Any ideas how to distinguish? @mzfr @rjt-gupta

tanner/emulators/twig_template_injection.py
```diff
@@ -0,0 +1,77 @@
...
+        return detection
+
+    async def handle(self, attack_params):
```
afeena, line 63: `session=None`

the signature of the function is wrong. base emulator calls every handle with session
`emulation_result = await self.emulators[detection['name']].handle(attack_params[detection['name']], session)`

tanner/emulators/twig_template_injection.py
rjt-gupta: For twig this case should give `6` and for tornado `222`.

rjt-gupta: we need something that distinguishes these two..

mzfr: yeah the regex is almost similar so it' hard to distinguish.

Also afeena can you please tell me how did you tested it? I mean the setup I want to know what I am doing wrong.

afeena: @mzfr scan works perfectly fine, if you print possible detection you can see it. the only problem in the signature of the `handle` function.

tanner/emulators/twig_template_injection.py
```diff
@@ -60,18 +60,7 @@ def scan(self, value):
...
-        result = await self.get_injection_result(attack_params[0]["value"])
-        if not result or "stdout" not in result:
-            self.logger.exception(
-                "Error while getting the injection results from php sandbox.."
-            )
-            return dict(status_code=504)
-        return dict(value=result["stdout"], page=False)
+    async def handle(self, attack_params, session=None):
+        attack_params[0]['value'] = unquote(attack_params[0]['value'])
+        result = await self.get_injection_result(attack_params[0]['value'])
```
rjt-gupta, line 65: tornado is getting its results from the custom docker image, and here its phpox. the returned format is different..

rjt-gupta (changes requested):

tanner/emulators/twig_template_injection.py
mzfr: @rjt-gupta do you have an example of the format?

PR #402, opened 2020-11-04
Fix migration script

mzfr:
@afeena 

Using zrevrange instead of zrange did give extra sessions but I am still not sure why this is happening. :sweat_smile:


afeena:
@mzfr looks good, but I will do additional testing as well once again to make sure everything is fine :)

PR #406, opened 2021-03-16
dl-2 domain not working

vncloudsco:
please check domain https://dl-2.alpinelinux.org/ not working. There is no need to change the domain, everything woking


vncloudsco:
Sorry for the mistake that the domain is still active, you can ignore it, but this domain doesn't support running https which makes the build unexpectedly error

afeena:
Thanks @vncloudsco !

Issue #423, opened 2022-06-21
Docker Installation of Tanner fails due to some dependencies


achute:
When I tried to install tanner using docker compose in Mac, I ran into the following python dependency issues:

```
tanner_web      | Traceback (most recent call last):
tanner_web      |   File "/opt/tanner/tanner-env/bin/tannerweb", line 4, in <module>
tanner_web      |     __import__('pkg_resources').run_script('Tanner==0.6.0', 'tannerweb')
tanner_web      |   File "/opt/tanner/tanner-env/lib/python3.9/site-packages/pkg_resources/__init__.py", line 651, in run_script
tanner_web      |     self.require(requires)[0].run_script(script_name, ns)
tanner_web      |   File "/opt/tanner/tanner-env/lib/python3.9/site-packages/pkg_resources/__init__.py", line 1455, in run_script
tanner_web      |     exec(script_code, namespace, namespace)
tanner_web      |   File "/opt/tanner/tanner-env/lib/python3.9/site-packages/Tanner-0.6.0-py3.9.egg/EGG-INFO/scripts/tannerweb", line 4, in <module>
tanner_web      |     __import__('pkg_resources').run_script('Tanner==0.6.0', 'tannerweb')
tanner_web      |   File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
tanner_web      |   File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
tanner_web      |   File "<frozen importlib._bootstrap>", line 664, in _load_unlocked
tanner_web      |   File "<frozen importlib._bootstrap>", line 627, in _load_backward_compatible
tanner_web      |   File "<frozen zipimport>", line 259, in load_module
tanner_web      |   File "/opt/tanner/tanner-env/lib/python3.9/site-packages/Tanner-0.6.0-py3.9.egg/tanner/web/server.py", line 3, in <module>
tanner_web      |   File "/opt/tanner/tanner-env/lib/python3.9/site-packages/aiohttp_jinja2/__init__.py", line 8, in <module>
tanner_web      |     from .helpers import GLOBAL_HELPERS
tanner_web      |   File "/opt/tanner/tanner-env/lib/python3.9/site-packages/aiohttp_jinja2/helpers.py", line 8, in <module>
tanner_web      |     @jinja2.contextfunction
tanner_web      | AttributeError: module 'jinja2' has no attribute 'contextfunction'
tanner_web exited with code 1

```

Jinja2 website says:

```
[...]

PR #424, opened 2022-06-28
Fix Jinja2 and MarkupSafe version issues

nikAizuddin:
Fixes https://github.com/mushorg/tanner/issues/423


rybolov:
This.  Worked like a charm as far as I can tell.

PR #433, opened 2024-03-18
'<3.0.0' is removed and dependency problem is resolved

### LINKED ISSUE #432: Please, remove "<3.0.0" from jinja paquet inside "requirements.txt" file
It gives a depencendy problem with aiohttp which is solved not specifying any version for jinja package.
Thanks!

saadali05:
Fixes #432


afeena:
Thank you for your contribution!

